PR #19, opened 2015-03-24
Use set to remove the duplicate candidates

kid1412z:
Hi Haifeng,
    Thank you for all your efforts. This commit change the way removing duplicated candidates, avoiding ordering, the running time of the toy data can be reduced almost half.
    By the way, I have been working on LSH based on SimHash or MinHash for weeks, and I'm wondering if I could contribute to this amazing project :~)
     Qiyang

### DIFF
--- a/Smile/src/main/java/smile/neighbor/LSH.java
+++ b/Smile/src/main/java/smile/neighbor/LSH.java
@@ -15,10 +15,7 @@
  *******************************************************************************/
 package smile.neighbor;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.*;
 import smile.math.IntArrayList;
 import smile.math.Math;
 import smile.sort.HeapSelect;
@@ -477,36 +474,13 @@ public class LSH <E> implements NearestNeighborSearch<double[], E>, KNNSearch<do
 
     @Override
     public Neighbor<double[], E> nearest(double[] q) {
-        IntArrayList candidates = new IntArrayList();
-
-        for (Hash h : hash) {
-            BucketEntry bucket = h.get(q);
-            if (bucket != null) {
-                int m = bucket.entry.size();
-                for (int i = 0; i < m; i++) {
-                    int index = bucket.entry.get(i);
-                    candidates.add(index);
-                }
-            }
-        }
-
+        Set<Integer> candidates = obtainCandidates(q);
         Neighbor<double[], E> neighbor = new Neighbor<double[], E>(null, null, -1, Double.MAX_VALUE);
-
-        int[] cand = candidates.toArray();
-        Arrays.sort(cand);
-        int prev = -1;
-        for (int index : cand) {
-            if (index == prev) {
-                continue;
-            } else {
-                prev = index;
-            }
-
+        for (int index : candidates) {
             double[] key = keys.get(index);
             if (q == key && identicalExcluded) {
                 continue;
             }
-
             double distance = Math.distance(q, key);
             if (distance < neighbor.distance) {
                 neighbor.index = index;
@@ -524,20 +498,7 @@ public class LSH <E> implements NearestNeighborSearch<double[], E>, KNNSearch<do
         if (k < 1) {
             throw new IllegalArgumentException("Invalid k: " + k);
         }
-
-        IntArrayList candidates = new IntArrayList();
-
-        for (Hash h : hash) {
-            BucketEntry bucket = h.get(q);
-            if (bucket != null) {
-                int m = bucket.entry.size();
-                for (int i = 0; i < m; i++) {
-                    int index = bucket.entry.get(i);
-                    candidates.add(index);
-                }
-            }
-        }
-
+        Set<Integer> candidates = obtainCandidates(q);
         Neighbor<double[], E> neighbor = new Neighbor<double[], E>(null, null, 0, Double.MAX_VALUE);
         @SuppressWarnings("unchecked")
         Neighbor<double[], E>[] neighbors = (Neighbor<double[], E>[]) java.lang.reflect.Array.newInstance(neighbor.getClass(), k);
@@ -546,18 +507,8 @@ public class LSH <E> implements NearestNeighborSearch<double[], E>, KNNSearch<do
             heap.add(neighbor);
         }
 
-        int[] cand = candidates.toArray();
-        Arrays.sort(cand);
-        
-        int prev = -1;
         int hit = 0;
-        for (int index : cand) {
-            if (index == prev) {
-                continue;
-            } else {
-                prev = index;
-            }
-
+        for (int index : candidates) {
             double[] key = keys.get(index);
             if (q == key && identicalExcluded) {
                 continue;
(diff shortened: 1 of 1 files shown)


Smile/src/main/java/smile/neighbor/LSH.java
```diff
@@ -15,10 +15,7 @@
  *******************************************************************************/
 package smile.neighbor;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.*;
```
haifengl: It is better not using * for import. Can you please import only necessary classes?

haifengl:
Thank you very much for the improvements! Have you run the unit test? The test uses USPS data, which you can easily find on internet.

You are very welcome to contribute to the project. If you can push your work on SimHash and MinHash, it will be great. BTW, I am not satisfied with my implementation of MPLSH. I will really appreciate if you improve it too. Thanks a lot!

Smile/src/main/java/smile/neighbor/LSH.java
kid1412z: I have some trouble running the test, when loading the dataset, it throws
exceptions:

java.lang.NumberFormatException: empty String
at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:1020)
at java.lang.Double.valueOf(Double.java:504)
at smile.data.NumericAttribute.valueOf(NumericAttribute.java:62)
at smile.data.parser.DelimitedTextParser.parse(DelimitedTextParser.java:304)
at smile.neighbor.LSHTest.<init>(LSHTest.java:46)
at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
at
sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
...

Maybe the problem is:

String[] s = line.split(delimiter, -1); //DelimitedTextParser.java line:250

it produced String "" but didn't ignore.

the data come from
http://statweb.stanford.edu/~tibs/ElemStatLearn/data.html and MD5
(zip.train) = a2d376ecc766edab27f0c4621f180f99  MD5 (zip.test) =
ec27c60f11cf4a80de34679dd9108b4d

And I'm following up this case.

About the import, my IDE import-optimizer do this automatically, I will fix it.




2015-03-24 10:21 GMT+08:00 Haifeng Li <notifications@github.com>:

> In Smile/src/main/java/smile/neighbor/LSH.java
> <https://github.com/haifengl/smile/pull/19#discussion_r26999133>:
>
> > @@ -15,10 +15,7 @@
> >   *******************************************************************************/
> >  package smile.neighbor;
> >
> > -import java.util.ArrayList;
> > -import java.util.Arrays;
> > -import java.util.LinkedList;
> > -import java.util.List;
> > +import java.util.*;
>
> It is better not using * for import. Can you please import only necessary
> classes?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/haifengl/smile/pull/19/files#r26999133>.
>



-- 

Zuo Qiyang (左启阳)

Beijing University of Posts and Telecommunications,

No.10 Xitucheng Rd., Haidian, Beijing, China 100876

中国, 北京市海淀区西土城路10号, 北京邮电大学, 100876

Mobile: +86  18810300667

kid1412z:
I have some trouble running the test, when loading the dataset, it throws
exceptions:

java.lang.NumberFormatException: empty String
at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:1020)
at java.lang.Double.valueOf(Double.java:504)
at smile.data.NumericAttribute.valueOf(NumericAttribute.java:62)
at smile.data.parser.DelimitedTextParser.parse(DelimitedTextParser.java:304)
at smile.neighbor.LSHTest.<init>(LSHTest.java:46)
at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
at
sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
...

Maybe the problem is:

String[] s = line.split(delimiter, -1); //DelimitedTextParser.java line:250

it produced String "" but didn't ignore.

the data come from
http://statweb.stanford.edu/~tibs/ElemStatLearn/data.html and MD5
(zip.train) = a2d376ecc766edab27f0c4621f180f99  MD5 (zip.test) =
ec27c60f11cf4a80de34679dd9108b4d

And I'm following up this case.

About the import, my IDE import-optimizer do this automatically, I will fix it.

haifengl:
Don't worry about it. I will pull your changes and do the tests. Thanks!

kid1412z:
Is it the right dataset?

haifengl:
It is the right dataset, but the format is different. In the LSHTest, add the following to the second line

parser.setDelimiter(",");

Thanks!

kid1412z:
I was confused, the data I have get is delimited by whitespace so I'd like to confirm it again, is it http://statweb.stanford.edu/~tibs/ElemStatLearn/datasets/zip.train.gz and http://statweb.stanford.edu/~tibs/ElemStatLearn/datasets/zip.test.gz ?

haifengl:
Sorry, I double checked and it is space delimited. I think that the problem is that there is extra white space at the end of each line. For now, you can just remove all the trailing space manually. It is easy to do in vim or an IDE.

For long term, we want to improve the parser. The problem is that it could introduce problems by trimming each line because sometime it is intended for missing values.

haifengl:
I tested the changes. On my machine, it is slightly faster in some cases, but slightly slower on other cases. It is hard to tell the difference. I wonder how you find that it is twice faster on toy data. Thanks!

But overall I like your changes, which is more concise. Thanks!

haifengl:
BTW, the recall rate matches previous values, which is good.

kid1412z:
I just run testToy(), and watch the printed time of NN, KNN and range and I was not rigorous without using the real test set.
About the data parser, do we allow using maven lib like guava? it provide like Splitter.on(',').omitEmptyStrings().split(str), or we will do it all by ourselves?

haifengl:
The problem is not abou space but missing values. Let's use _ as space.

1__1

Means 1,?,1 where ? Stands for missing value. A trailing space may indicate a missing value.

haifengl:
I pull the merge because the code is cleaner. Thanks! Would you like to look into mplsh? The performance is not good so far. 

I will look into the parser again.

kid1412z:
OK, I willl check it out.

haifengl:
String[] s = line.split(delimiter, -1); //DelimitedTextParser.java line:250

Can you replace -1 with 0 and then try your data? I am afraid that this change will break other cases. So please just change it in your local branch right now. Thanks!

haifengl:
Thank you very much for looking into mplsh!

kid1412z:
I did not change the unit test, "String[] s = line.split(delimiter, -1);"  is the origin one.   I still have problem running the test. And I can find _ or ? in my dataset.... :(

haifengl:
I mean change -1 to 0 and try again. I just use ? and _ for an example, not about your data.

kid1412z:
Still fail: java.text.ParseException: 258 columns, expected 257.

haifengl:
Sorry. You also need change line 362

kid1412z:
It works, thanks!

PR #27, opened 2015-04-15
set threads as daemons

guillaumecherel:
The threads created by the MulticoreExecutor prevented the JVM from exiting after the computation is finished. I use smile from scala using sbt, and this prevented the computation from exiting cleanly while leaving sbt running. I had to kill the JVM to terminate the program (including sbt which runs in the same JVM as the code). Setting the threads as daemons solved the problem.


haifengl:
This is great! Thanks! BTW, you can call MulticoreExecutor.shutdown to kill the thread pool.

guillaumecherel:
Oh right, thanks. I hadn't seen the shutdown method.

Glad I could help!

Issue #32, opened 2015-05-12
Is there an easy way to calculate multiclass precision, recall and F-Measure validation?


owlmsj:
Given two vectors, the predicted and the truth, filled with multi class numbers is there an easy way to computate precision, recall and f-measure?

Example:

predicted = [0,2,1,3]
truth         = [0,1,2,3]

haifengl:
Check out the package smile.validation. We have all of them (and something else) there.

owlmsj:
Sure, but in Precision, Recall and FMeasure implementation is only for one-class validation, right? It test 0 (false) and 1 (true), not the class-number. Or what am I doing wrong?

owlmsj:
This question is like in this link : http://stats.stackexchange.com/questions/51296/how-to-calculate-precision-and-recall-for-multiclass-classification-using-confus

haifengl:
Ok. Let's create a class to calculate confusion matrix and then update these measure classes for multi-class.

owlmsj:
Hi, I did that in meantime. Any suggestions about this new class name?

haifengl:
ConfusionMatrix should be fine. You can make a pull request and we can start from there. Thanks!

owlmsj:
Okay! Just a little con: Project 'Smile' is missing required Java project: 'SmileTestData'	. Where can I find this project?

haifengl:
SmileTestData contains various datasets that I download from internet. As I am not the owner of these datasets, I cannot include them in github.

owlmsj:
Ah, no problem. I just remove the dependency and it was OK!

haifengl:
The maven .pom files have no such dependency. Only for eclipse project files. BTW, SmileTestData is only for unit tests. You don't need them to compile the packages. Thanks!

haifengl:
Thanks for the great work! The ConfusionMatrix is merged into master.

PR #34, opened 2015-05-18
Confusion Matrix implementation to compute multi-class precision, recall and f-measure

owlmsj:
finally, pull request from issue #32

### DIFF
--- a/Smile/src/main/java/smile/validation/ConfusionMatrix.java
+++ b/Smile/src/main/java/smile/validation/ConfusionMatrix.java
@@ -1,0 +1,60 @@
+package smile.validation;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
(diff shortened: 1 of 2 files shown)


Smile/src/main/java/smile/validation/ConfusionMatrix.java
```diff
@@ -0,0 +1,60 @@
...
+public class ConfusionMatrix {
+	
+	private int[][] matrix;
+	
+	
+	public int[][] getMatrix(){
+		return matrix;
+	}
+	
+	public int[][] generate(int[] truth, int[] prediction) {
```
haifengl: Nice work. Shall we just change generate to the constructor. It is not necessary to create an object and then call this method. Let's do it in one step. Thanks!

PR #45, opened 2015-11-06
Build was failing with maven 3.3.3 on JDK7

chhh:
The reason was -Xdoclint:none specified in javadoc plugin was not recognized.
The problem should be solved by adding a profile which only adds the switch if JDK >= 1.8 is detected.

Also added /target directories to .gitignore.


haifengl:
Thank you!

PR #49, opened 2015-12-05
Fixed a bug in DecisionTree that falseCount is improperly shared

myui:
I think `falseCount` is improperly shared when Exception caused in 
https://github.com/haifengl/smile/blob/master/Smile/src/main/java/smile/classification/DecisionTree.java#L431

`falseCount` is unique for each invocation of `findBestSplit` in 
https://github.com/haifengl/smile/blob/master/Smile/src/main/java/smile/classification/DecisionTree.java#L478

So, it would be better to change `falseCount` a local variable of `findBestSplit`.


myui:
Ah, sorry it was okey to be shared for serial execution. I'll close this PR.

PR #52, opened 2015-12-28
Fix ArrayIndexOutOfBoundsException when accessing math util's random.

jamespedwards42:
Access to the utility Random was synchronized on a method level instead
of an object level.

Not sure if this is how you want to fix it, but I was running into this Exception when I had multiple threads learning/clustering/etc. across multiple core objects that use the Math util internally.  I just went with the easiest fix.

### DIFF
--- a/math/src/main/java/smile/math/Math.java
+++ b/math/src/main/java/smile/math/Math.java
@@ -786,8 +786,10 @@ public class Math {
      * need to generate pseudorandom numbers at a great rate, it may reduce
      * contention for each thread to have its own pseudorandom-number generator. 
      */
-    public static synchronized double random() {
-        return random.nextDouble();
+    public static double random() {
+        synchronized(random) {
+            return random.nextDouble();
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


haifengl:
Thanks for fix. Because these methods are static and I assume synch on method should have same effect (it actually syncs on the class object Math). I will research why it causes the problem.

BTW, I usually create a Random object in parallel algorithms that need random numbers, which avoids the problem and also improves the performance. Can you share your use case that multiple threads learning/clustering? If it is important to you, I will use thread load variable to avoid sync. Thanks!

haifengl:
I make the random variable thread local. It should be more efficient and safe. Thanks!

jamespedwards42:
Nice, yeah, a ThreadLocal is way better.  

One use case is for parallelizing cross validation. And then I also have another couple of threads clustering using NeuralGas to select data sets similar to what I'm trying to predict.  Specifically I've been using the NeuralNetwork, KNN and NeuralGas classes that all were hitting that Random.

haifengl:
Interestingly. I guess that you also use other machine learning tools. How do you feel SMILE? Anything you want us to improve? Compared to others, is it fast enough? Thanks!

jamespedwards42:
SMILE is amazing!!! Thank you so much for creating this, it is so nice to have such a clean interface into everything.  Sorry, no reference point on speed, but I think that issue usually needs to be solved on a horizontal scale, so the real difference maker for me is the ability to write clean/maintainable/reusable code against your framework/library. I'm a complete beginner when it comes to machine learning, just a traditional CS background.  I feel like this library will really help open doors for people to start learning about this field.

haifengl:
Thanks! I am working on a shell for SMILE. It is embedded into scala and provides a much simpler api for interactive use.

Issue #54, opened 2015-12-29
Handling of identical objects in DBScan


badgersow:
Hi,

First let me thank you for the great library, it is pleasure to use it.

In my work project I use DBScan class to cluster a bag of strings.
I have noticed that by default identical objects are ignored and not clustered together.

Please consider the following test:
``` java
@Test
public void testIdentical() {
    String[] dataset = new String[]{"a", "a", "a", "a", "b"};
    Distance<String> distance = new Distance<String>() {
        @Override
        public double d(String x, String y) {
            return x.equals(y) ? 0.0 : 1.0;
        }
    };

    DBScan<String> clusterer = new DBScan<>(dataset, distance, 2, 0.5);
    int[] labels = clusterer.getClusterLabel();
    assertTrue(labels[0] != DBScan.OUTLIER);
}
```

It is failing because strings "a" are not clustered together and are considered as outliers.
I believe this behavior is not intuitive.
The problem is not critical and I managed to workaround this with the following lines:
``` java
LinearSearch<String> search = new LinearSearch<>(allNames, JARO_WINKLER_DISTANCE);
search.setIdenticalExcluded(false);
DBScan<String> clusterer = new DBScan<>(allNames, search, MIN_PTS, RADIUS);
```
But nevertheless, do you think the default behavior is ok?
Should we at least provide constructor to DBScan to allow to create model with identical objects not excluded?

Thanks for your time.

haifengl:
Thanks! As you may figure it out yourself, DBSCan use RNNSearch inside. When a Distance or Metric is supplied, we wrap it with LinearSearch or CoverTree. setIdenticalExcluded(true) is important for some uses including DBScan. But for your use case, it is different due to the meaning of JAR_WINKLER_DISTANCE. I will check if I should update  the semantical meaning of setIdenticalExcluded, for example it means exactly same object (not just distance 0). I have to double check before making this change. Thanks!

haifengl:
BTW, can you please contribute your JARO_WINKLER_DISTANCE implementation to SMILE? Thank you very much in advance! If you agree, please put it in package smile.math.distance. Thanks again!

haifengl:
[...]

PR #55, opened 2015-12-29
Clarify documentation on neighborhood search.

badgersow:
Summarize issue #54 in documentation


haifengl:
setIdenticalExclude is really an implementation feature, not at this interface level. Actually, different nearest neighbor search data structures and algorithms may have different behaviors. I will clarify this in each implementation. Thanks!

haifengl:
I add details comments on setIdenticalExcluded for all nearest neighbor search classes. Thanks!

PR #63, opened 2016-01-20
Change library versions to numbers that actually exist in Sonatype.

ejwu:
The SBT command as written doesn't work.  Based on the Sonatype link it seems like the highest version is actually 1.0.4, instead of 1.1.0.

Note that I don't use Maven so I haven't tested that change, but the SBT dependency appears to work now.

### DIFF
--- a/shell/src/universal/doc/faq.html
+++ b/shell/src/universal/doc/faq.html
@@ -20,21 +20,21 @@
     <h1 id="top" class="title">Smile FAQ</h1>
     <h2 class="question" id="link-with-smile">Link with Smile</h2>
     <p class="answer">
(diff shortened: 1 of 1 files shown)


haifengl:
The website is companioned with 1.1.0 because scala api is only available from 1.1.0. The version 1.1.0 is still in rapid development and the website is not really open to public yet. Sorry for confusion. As you can tell, the website is still in construction. I will remove the link in readme.md. 

BTW, I guess that you are using scala. If so, I strongly recommend you try the new scala api in the shell. In your clone, just type ./smile.sh, which will build and start the shell. Thanks!

PR #69, opened 2016-02-28
GMeans: add constructor parameters to omit terminal output and allow to set the number of maximal iterations

christianolms:
I have added two constructor parameters to allow a caller to omit the Console Output (to be able to use SMILE for a project where I need a clean terminal output) and to be able to set the maxIter parameter by the creation of the Kmeans instances.

### DIFF
--- a/core/src/main/java/smile/clustering/GMeans.java
+++ b/core/src/main/java/smile/clustering/GMeans.java
@@ -48,6 +48,16 @@ public class GMeans extends KMeans {
      * @param kmax the maximum number of clusters.
      */
     public GMeans(double[][] data, int kmax) {
+    	this(data,kmax,100,true);
+    }
+    /**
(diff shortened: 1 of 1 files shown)


haifengl:
There are many algorithms printing out progress information. How about changing them to sl4j systematically?

haifengl:
I have updated all algorithms using SLF4J for progress, debug, and error messages. In your applications, if you don't include any SLF4J bindings (e.g. log4j, logback, etc.), these will be nop (nothing will be printed). If you include some bindings but don't want see these message, please adjust the log level above INFO. Thanks!

christianolms:
Dear haifengl,
thank you very much for the fast reply. The solution with mapping the algorithm progress to SLF4J is a much better approach than my suggestion. I will even go further and think of adding SLF4J to my own project for logging. Thank you for the recommendation.

PR #80, opened 2016-05-20
Make Simple and PennTreebank tokenizers Unicode-aware

arehart13:
Hi Haifeng,
The SimpleTokenizer and PennTreebankTokenizer rely on the regex character class `\w` in the first delimiter pattern. Unfortunately, `\w` only matches the ASCII letters and numbers, so words with diacritized characters or non-Latin characters are not recognized as tokens. For example, `résumé` is tokenized as `["r", "é", "sum", "é"]`. 

I replaced `\w` with the following: `\p{L}\p{N}\p{M}\p{Sk}\u200C\u200D`. This expression matches the unicode letter category (L), the number category (N), the marks category (M, which includes combining diacritics), some miscellaneous diacritics (Sk), and the zero-width non-joiner and joiner characters (used word-internally mostly in Arabic script but sometimes in other languages).

These changes should improve tokenization on English and European languages, as well as providing a decent baseline tokenization for non-Western whitespace-delimited languages.

### DIFF
--- a/nlp/src/main/java/smile/nlp/tokenizer/PennTreebankTokenizer.java
+++ b/nlp/src/main/java/smile/nlp/tokenizer/PennTreebankTokenizer.java
@@ -68,7 +68,7 @@ public class PennTreebankTokenizer implements Tokenizer {
 
     private static final Pattern[] DELIMITERS = {
         // Separate most punctuation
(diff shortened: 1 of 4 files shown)


haifengl:
Thank you very much! This is really nice!

PR #81, opened 2016-05-20
Use the UNICODE_CHARACTER_CLASS flag (?U) for both \s and \w in token…

arehart13:
I ran into another issue with the tokenizers: not recognizing non-ascii space chars such as the no-break space (often found in web pages) and em-space. I believe the more general (and correct) solution is to simply enable the UNICODE_CHARACTER_CLASS flag on the tokenizer regexes. This also simplifies the previous change since `\w` will just do the right thing now.


haifengl:
Cool. Thanks!

PR #83, opened 2016-05-23
Text normalization

arehart13:
Haifeng,
While using Smile's tokenizer and pos-tagger, I came across a need to perform basic text normalization. I added a normalization interface and a simple normalizer that applies the Unicode normalization form NFKC, trims/compresses/normalizes whitespace, removes formatting and control characters, and replaces non-ASCII double and single quotes (i.e. "curly quotes") with their ASCII versions. I hope others might find it useful.


haifengl:
Thank you very much! This is awesome!

PR #93, opened 2016-06-20
Fix: squid:S1210, "equals(Object obj)" should be overridden along wit…

elkfrawy-df:
 squid:S1210- “""equals(Object obj)"" should be overridden along with the ""compareTo(T obj)"" method”. 
 https://dev.eclipse.org/sonar/rules/show/squid:S1210

Ayman Elkfrawy

### DIFF
--- a/core/src/main/java/smile/classification/DecisionTree.java
+++ b/core/src/main/java/smile/classification/DecisionTree.java
@@ -409,6 +409,26 @@ public class DecisionTree implements SoftClassifier<double[]>, Serializable {
             return (int) Math.signum(a.node.splitScore - node.splitScore);
         }
 
+        @Override
+        public boolean equals(Object obj) {
+            if (obj == this) {
+                return true;
+            }
+            if (obj == null) {
+                return true;
+            }
+            if (obj instanceof TrainNode) {
+                return this.node.splitFeature == ((TrainNode)obj).node.splitScore;
+            }
+            return false;
+        }
+
+        @Override
+        public int hashCode() {
+            long fitnessLong = Double.doubleToLongBits(node.splitScore);
+            return 41 + (int) (fitnessLong ^ (fitnessLong >>> 32));
+        }
+
         /**
          * Task to find the best split cutoff for attribute j at the current node.
          */
--- a/core/src/main/java/smile/classification/DecisionTree.java
+++ b/core/src/main/java/smile/classification/DecisionTree.java
@@ -15,19 +15,19 @@
  *******************************************************************************/
 package smile.classification;
 
-import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.PriorityQueue;
-import java.util.concurrent.Callable;
-
 import smile.data.Attribute;
 import smile.data.NominalAttribute;
 import smile.data.NumericAttribute;
 import smile.math.Math;
 import smile.sort.QuickSort;
 import smile.util.MulticoreExecutor;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.PriorityQueue;
+import java.util.concurrent.Callable;
 
 /**
  * Decision tree for classification. A decision tree can be learned by
@@ -415,7 +415,7 @@ public class DecisionTree implements SoftClassifier<double[]>, Serializable {
                 return true;
             }
             if (obj == null) {
-                return true;
+                return false;
             }
             if (obj instanceof TrainNode) {
                 return this.node.splitFeature == ((TrainNode)obj).node.splitScore;
--- a/core/src/main/java/smile/gap/BitString.java
+++ b/core/src/main/java/smile/gap/BitString.java
@@ -199,7 +199,7 @@ public class BitString implements Chromosome {
             return true;
         }
         if (obj == null) {
-            return true;
+            return false;
         }
         if (obj instanceof Chromosome) {
             return this.fitness == ((Chromosome)obj).fitness();
--- a/core/src/main/java/smile/gap/BitString.java
+++ b/core/src/main/java/smile/gap/BitString.java
@@ -192,7 +192,27 @@ public class BitString implements Chromosome {
     public int compareTo(Chromosome o) {
         return (int) Math.signum(fitness - o.fitness());
     }
-    
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        }
+        if (obj == null) {
+            return true;
+        }
+        if (obj instanceof Chromosome) {
+            return this.fitness == ((Chromosome)obj).fitness();
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        long fitnessLong = Double.doubleToLongBits(fitness);
+        return 37 + (int) (fitnessLong ^ (fitnessLong >>> 32));
+    }
+
     @Override
     public double fitness() {
         if (Double.isNaN(fitness)) {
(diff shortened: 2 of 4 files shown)


core/src/main/java/smile/gap/BitString.java
```diff
@@ -192,7 +192,27 @@ public BitString(int[] bits, FitnessMeasure<BitString> measure, Crossover crosso
...
         return (int) Math.signum(fitness - o.fitness());
     }
-    
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        }
+        if (obj == null) {
```
haifengl: If obj is null, shall we return false?

core/src/main/java/smile/neighbor/CoverTree.java
```diff
@@ -202,6 +202,27 @@ E getObject() {
         public int compareTo(DistanceNode o) {
             return (int) Math.signum(dist - o.dist);
         }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (obj == this) {
+                return true;
+            }
+            if (obj == null) {
```
haifengl: same question as above

core/src/main/java/smile/neighbor/CoverTree.java
```diff
@@ -202,6 +202,27 @@ E getObject() {
...
+                return true;
+            }
+            if (this.getClass() == obj.getClass()) {
```
haifengl: For sake of consistency, shall we use instanceOf pattern as above/BiString?

core/src/main/java/smile/gap/BitString.java
elkfrawy-df: Yes, you are right, I will fix this.

core/src/main/java/smile/neighbor/MPLSH.java
```diff
@@ -228,6 +228,26 @@ public int compareTo(PrH o) {
             // to sort PrH in decreasing order.
             return (int) Math.signum(o.pr - pr);
         }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (obj == this) {
+                return true;
+            }
+            if (obj == null) {
```
haifengl: same issue with null

core/src/main/java/smile/neighbor/MPLSH.java
```diff
@@ -250,6 +270,29 @@ public int compareTo(PrZ o) {
             // to sort PrZ in decreasing order.
             return prh[0].compareTo(o.prh[0]);
         }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (obj == this) {
+                return true;
+            }
+            if (obj == null) {
```
haifengl: same issue with null

core/src/main/java/smile/neighbor/MPLSH.java
```diff
@@ -660,6 +703,26 @@ public int compareTo(Probe o) {
             return (int) Math.signum(prob - o.prob);
         }
 
+        @Override
+        public boolean equals(Object obj) {
+            if (obj == this) {
+                return true;
+            }
+            if (obj == null) {
```
haifengl: same issue with null

core/src/main/java/smile/neighbor/MPLSH.java
```diff
@@ -660,6 +703,26 @@ public int compareTo(Probe o) {
...
+                return true;
+            }
+            if (this.getClass() == obj.getClass()) {
```
haifengl: instanceOf?

core/src/main/java/smile/neighbor/CoverTree.java
elkfrawy-df: Actually that rises a compile time exception "Illegal generic type for instanceof", after I searched the problem, this was one of the solutions, the other solution is to make the class "DistanceNode" static. Should I make it static?

haifengl: Let's keep it as is and use this.getClass() == obj.getClass(). Thanks!

elkfrawy-df:
Please let me know if you want me to quash commits.

core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -15,20 +15,20 @@
...
-import java.util.concurrent.Callable;
-
 import smile.data.Attribute;
 import smile.data.NominalAttribute;
 import smile.data.NumericAttribute;
 import smile.math.Math;
 import smile.sort.QuickSort;
 import smile.util.MulticoreExecutor;
 
+import java.io.Serializable;
```
haifengl: I prefer java system package imports at the beginning. Thanks!

core/src/main/java/smile/neighbor/MPLSH.java
```diff
@@ -660,6 +704,26 @@ public int compareTo(Probe o) {
...
+                return this.prob == ((Probe)obj).prob;
+            }
+            return false;
+        }
+
+        @Override
+        public int hashCode() {
+            long fitnessLong = Double.doubleToLongBits(prob);
```
haifengl: why call it fitnessLong?

core/src/main/java/smile/neighbor/MPLSH.java
```diff
@@ -228,6 +229,26 @@ public int compareTo(PrH o) {
...
+            }
+            if (obj instanceof PrH) {
+                return this.pr == ((PrH)obj).pr;
+            }
+            return false;
+        }
+
+        @Override
+        public int hashCode() {
+            long fitnessLong = Double.doubleToLongBits(pr);
```
haifengl: why call it fitnessLong?

core/src/main/java/smile/neighbor/CoverTree.java
```diff
@@ -202,6 +203,27 @@ E getObject() {
...
+                return this.dist == ((DistanceNode)obj).dist;
+            }
+            return false;
+        }
+
+        @Override
+        public int hashCode() {
+            long fitnessLong = Double.doubleToLongBits(dist);
```
haifengl: why call it fitnessLong?

core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -409,6 +409,26 @@ public int compareTo(TrainNode a) {
...
+            }
+            if (obj instanceof TrainNode) {
+                return this.node.splitFeature == ((TrainNode)obj).node.splitScore;
+            }
+            return false;
+        }
+
+        @Override
+        public int hashCode() {
+            long fitnessLong = Double.doubleToLongBits(node.splitScore);
```
haifengl: why call it fitnessLong?

haifengl:
many added equals and hashCode are for internal classes. They never get chanced to call.

haifengl:
many hashCode implementation are copied from BitString and don't seem correct for those classes (missing other fields).

core/src/main/java/smile/gap/BitString.java
```diff
@@ -192,7 +192,27 @@ public BitString(int[] bits, FitnessMeasure<BitString> measure, Crossover crosso
...
+            return false;
+        }
+        if (obj instanceof Chromosome) {
+            return this.fitness == ((Chromosome)obj).fitness();
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
```
haifengl: this doesn't make sense. all other fields are ignored. and I don't expect BitString to be used in a hash map.

haifengl:
I think we should squash this request. Thanks!

elkfrawy-df:
Yes, sorry about messy copy-paste code, I will fix the names. Generally implementation of **equals** and **hashcode** are working on the same fields. So I looked at what you already implemented on **compareTo** and did the same on **equals** and **hashcode**.

haifengl:
compareTo is for sorting purpose, which is very important for these algorithms. But we never call equals and hasCode. The rule squid:S1210 actually doesn't apply here.

elkfrawy-df:
No problem then, we can close this pull request. 
Thanks for your respond.

haifengl:
Thank you very much for helping!!

haifengl:
BTW, are these all problems found your tools? Thanks!

elkfrawy-df:
No, there still much to fix, I just started with these two and I needed that response to continue on the rest.

haifengl:
Cool! Thanks a lot!

PR #96, opened 2016-06-23
Fix: squid:S2259, Null pointers should not be dereferenced

elkfrawy-df:
squid:S2259 - Null pointers should not be dereferenced
https://dev.eclipse.org/sonar/rules/show/squid:S2259

Ayman Elkfrawy.

### DIFF
--- a/plot/src/main/java/smile/swing/table/FontCellRenderer.java
+++ b/plot/src/main/java/smile/swing/table/FontCellRenderer.java
@@ -15,9 +15,8 @@
  *******************************************************************************/
 package smile.swing.table;
 
-import java.awt.Font;
-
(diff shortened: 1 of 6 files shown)


plot/src/main/java/smile/swing/table/FontCellRenderer.java
```diff
@@ -15,9 +15,8 @@
  *******************************************************************************/
 package smile.swing.table;
 
-import java.awt.Font;
-
 import javax.swing.table.DefaultTableCellRenderer;
+import java.awt.*;
```
haifengl: Can you change back to original way?

import java.awt.Font;

PR #97, opened 2016-06-23
Fix: squid:S1192, String literals should not be duplicated

elkfrawy-df:
 squid:S1192 - “String literals should not be duplicated”. 
 https://dev.eclipse.org/sonar/rules/show/squid:S1192

Ayman Elkfrawy.

### DIFF
--- a/core/src/main/java/smile/classification/AdaBoost.java
+++ b/core/src/main/java/smile/classification/AdaBoost.java
@@ -15,14 +15,14 @@
  *******************************************************************************/
 
 package smile.classification;
-
 import java.io.Serializable;
 import java.util.Arrays;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import smile.math.Math;
 import smile.data.Attribute;
 import smile.data.NumericAttribute;
+import smile.math.Math;
 import smile.util.SmileUtils;
 import smile.validation.Accuracy;
 import smile.validation.ClassificationMeasure;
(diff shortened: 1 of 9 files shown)


douglasArantes:
I think `static final String INVALID NUMBER OF TREES` must have the `private` modifier instead of public.

elkfrawy-df:
So should I make all `static final String` I added to be `private`? I think it would be better.

douglasArantes:
That's it.

haifengl:
Thanks! More importantly, we should use -XX:+UseStringDeduplication for JVM, which is more important for runtime. I don't think that we will have millions smile model in an application. So the duplicated message strings are big issues. But when we parse data files, we usually get a lot of duplicated string tokens. So -XX:+UseStringDeduplication is very important in that use case.

BTW, is it your lint tools or IDE reorganize the import statements? I see that it always move java system import to the bottom of imports and use .*, which I don't like.

Sometimes, it is dangerous to move java system import to the end. For example, we may have same class names (e.g. smile.math.Math). Can you please turn of the import reordering and .*? Thanks!

elkfrawy-df:
Sorry about the import problem, I modified it no to do so anymore. I will fix all imports in my pull requests.
About the string duplication, these fixes are not meant only for performance matters, it is most intended to improve maintainability. 
> constants can be referenced from many places, but only need to be updated in a single place.

haifengl:
Thanks for updates. Can you please don't use things like

import java.awt.*

Things can be weird in case of conflict name space such as java.awt.Event and com.mycompany.calendar.Event.

As you can see, I never use import wildcard in my code. Thanks!

elkfrawy-df:
@haifengl I will squash and resolve the conflicts

haifengl:
@elkfrawy-df the conflict is a simple one line change (some typo fix). Can you please resolve it and I will merge the request. Thanks!

elkfrawy-df:
@haifengl Sorry, I forgot to resolve it, I will now.

PR #98, opened 2016-06-23
Fix: squid:S1854, Dead stores should be removed

elkfrawy-df:
 squid:S1854 - “ Dead stores should be removed ”. 
 https://dev.eclipse.org/sonar/rules/show/squid:S1854

Ayman Elkfrawy.

### DIFF
--- a/core/src/main/java/smile/clustering/CLARANS.java
+++ b/core/src/main/java/smile/clustering/CLARANS.java
@@ -230,7 +230,7 @@ public class CLARANS <T> extends PartitionClustering<T> {
         int n = data.length;
 
         int index = Math.randomInt(k);
-        T medoid = null;
+        T medoid;
         boolean dup;
         do {
             dup = false;
--- a/data/src/main/java/smile/data/parser/ArffParser.java
+++ b/data/src/main/java/smile/data/parser/ArffParser.java
@@ -183,7 +183,7 @@ public class ArffParser {
      */
     private String readHeader(StreamTokenizer tokenizer, List<Attribute> attributes) throws IOException, ParseException {
         /// The name of dataset.
-        String relationName = null;
+        String relationName;
         // clear attribute set, which may be from previous parsing of other datasets.
         attributes.clear();
 
--- a/graph/src/test/java/smile/graph/AdjacencyListTest.java
+++ b/graph/src/test/java/smile/graph/AdjacencyListTest.java
@@ -349,8 +349,6 @@ public class AdjacencyListTest {
     @Test
     public void testDfs2() {
         System.out.println("dfs connected component");
-        int[] size = {3, 5};
-        int[] id = {0, 1, 0, 1, 1, 1, 0, 1};
         int[][] cc = {{0, 2, 6}, {1, 3, 4, 5, 7}};
 
         Graph graph = new AdjacencyList(8);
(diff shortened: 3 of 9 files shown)


core/src/main/java/smile/clustering/CLARANS.java
```diff
@@ -230,7 +230,7 @@ private double getRandomNeighbor(T[] data, T[] medoids, int[] y, double[] d) {
         int n = data.length;
 
         int index = Math.randomInt(k);
-        T medoid = null;
+        T medoid;
```
haifengl: This is actually to remove warning like "uninitialized variable". Maybe today javac doesn't report that warning any more?

data/src/main/java/smile/data/parser/ArffParser.java
```diff
@@ -183,7 +183,7 @@ private void getNextToken(StreamTokenizer tokenizer) throws IOException, ParseEx
      */
     private String readHeader(StreamTokenizer tokenizer, List<Attribute> attributes) throws IOException, ParseException {
         /// The name of dataset.
-        String relationName = null;
+        String relationName;
```
haifengl: same as above

graph/src/test/java/smile/graph/AdjacencyListTest.java
```diff
@@ -349,8 +349,6 @@ public void testDfs() {
     @Test
     public void testDfs2() {
         System.out.println("dfs connected component");
```
haifengl: Please keep these variables. I may miss some test on these results. I should add more tests, not remove them. Thanks!

graph/src/test/java/smile/graph/AdjacencyListTest.java
```diff
@@ -423,8 +421,6 @@ public void testBfs() {
     @Test
     public void testBfs2() {
         System.out.println("bfs connected component");
-        int[] size = {3, 5};
```
haifengl: same as above

graph/src/test/java/smile/graph/AdjacencyMatrixTest.java
```diff
@@ -348,8 +348,6 @@ public void testDfs() {
     @Test
     public void testDfs2() {
         System.out.println("dfs connected component");
-        int[] size = {3, 5};
```
haifengl: same as above

haifengl:
I think we should withdraw this one. These dead stores are not really errors and have no performance hit. They were used on purpose before to make javac happy. Otherwise, we will get a lot of warnings.

elkfrawy-df:
@haifengl This is more for readability purposes, that there is no meaning for the initial value and it might confuse the developer.

plot/src/main/java/smile/swing/FontChooser.java
```diff
@@ -484,7 +483,7 @@ public TextFieldKeyHandlerForListSelectionUpDown(JList<String> list) {
 
         @Override
         public void keyPressed(KeyEvent e) {
-            int i = targetList.getSelectedIndex();
+            int i;
             switch (e.getKeyCode()) {
```
elkfrawy-df: This assignment for example is redundant and has no effect which reduces the readability.

haifengl:
Without them, I will get a lot of warnings. Have you check build.sbt on javac options. I guess that your compiler settings are different.

haifengl:
I updated FontChooser by removing i. Please have a look.

haifengl:
Please remove the change to FontChooser. I will merge the rest changes. Thanks!

haifengl:
I understand these assignments are useless in Java because the language does it implicitly. However, I prefer to keep them in case that I may port the code to c/c++ and/or cuda. If so, I can simply copy the code without worries I miss initializations.

elkfrawy-df:
That is fine, no problem.

PR #99, opened 2016-06-23
Fix: squid:S1118, Utility classes should not have public constructors

elkfrawy-df:
 squid:S1118 - “Utility classes should not have public constructors ”. 
 https://dev.eclipse.org/sonar/rules/show/squid:S1118

Ayman Elkfrawy.

### DIFF
--- a/core/src/main/java/smile/neighbor/SNLSH.java
+++ b/core/src/main/java/smile/neighbor/SNLSH.java
@@ -214,6 +214,9 @@ public class SNLSH<E> implements NearestNeighborSearch<SNLSH.AbstractSentence, E
     public static class SimHash {
         private static final long seed = 0; //do not change seed
 
+        private SimHash() {
+        }
+
         public static long simhash64(List<String> tokens) {
             final int BITS = 64;
             if (tokens == null || tokens.isEmpty()) {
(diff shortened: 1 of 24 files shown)


haifengl:
Many changes here are wrong. For example, for all validation and hash functions, we should be able to create instances and pass to other functions, which we do use in practice.

elkfrawy-df:
@haifengl I can't see any good use of that. Why would you create an object of such utility class and pass it around?

haifengl:
For model validation, we have several `test` functions in scala that take model validation objects. It is up to the users to decide which measure to use.

haifengl:
Similarly for all classes for sort, BIC, etc. We provide static functions so that people can call them directly without create an object. But it is also possible for user to decide which measure to use. For example, in a GUI/web application, the user can drag and drop measures to test model. In that case, we need to create objects.

elkfrawy-df:
Well if you have specific usage for such classes that is fine. But as a general practice, it is better for utility classes to be passed as class not as an object, specially there is no any polymorphism or inheritance there.

haifengl:
I think that many classes are more complicated than adding a private constructor. There are more space to improve. For example, we can change the methods of MulticoreExecutor.java to not static. Then we can have multiple executors in runtime, which may be useful in a large system.

Instead of merging the whole request, I will review each case, cherry pick some of your changes and manually update the code. Thanks!

haifengl:
I cherry pick your changes. Thank you very much!

PR #112, opened 2016-08-23
Supported partial decomposition

myui:
Partial decomposition is sometimes useful. 

So, this PR supports partial decomposition of CholeskyDecomposition as seen in  
http://math.nist.gov/javanumerics/jama/doc/Jama/CholeskyDecomposition.html
http://grepcode.com/file_/repo1.maven.org/maven2/gov.nist.math.jama/gov.nist.math.jama/1.1.0/Jama/CholeskyDecomposition.java/?v=source

`isspd = isspd & (A[k][j] == A[j][k]);` is not performed since smile assumes it.

### DIFF
--- a/math/src/main/java/smile/math/matrix/CholeskyDecomposition.java
+++ b/math/src/main/java/smile/math/matrix/CholeskyDecomposition.java
@@ -55,10 +55,17 @@ public class CholeskyDecomposition {
     private double[][] L;
 
     /**
+     * Symmetric and positive definite flag.
+     * 
+     * @serial is symmetric and positive definite flag.
+     */
+    private final boolean isSPD;
+    
+    /**
      * Constructor. Used by newInstance().
      */
     private CholeskyDecomposition() {
-        
+        isSPD = true;
     }
     
     /**
@@ -83,6 +90,10 @@ public class CholeskyDecomposition {
     public CholeskyDecomposition(double[][] A) {
         this(A, false);
     }
+    
+    public CholeskyDecomposition(double[][] A, boolean overwrite) {
+        this(A, overwrite, false);
+    }
 
     /**
      * Constructor. Cholesky decomposition for symmetric and positive definite matrix. The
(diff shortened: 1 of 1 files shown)


haifengl:
Why do you want to change some variables and init to protected in multi variate gaussian?

haifengl:
What's the use case of partial decomposition?

myui:
@haifengl (cc: @L3Sota)

> Why do you want to change some variables and init to protected in multi variate gaussian?

Reverted the change about multi-variate gaussian for the time being.

> What's the use case of partial decomposition?

My use case is computing the inverse of sigma in the multi variate gaussian as seen in 
https://github.com/L3Sota/hivemall/blob/36990d10e1ebae61604a3cfa2b808577568c72cd/core/src/main/java/hivemall/utils/math/MultivariateGaussianDistribution.java#L169
https://github.com/L3Sota/hivemall/blob/36990d10e1ebae61604a3cfa2b808577568c72cd/core/src/main/java/hivemall/anomaly/ChangeFinderUDF.java#L469 (Autoregression-like algorithm named SDAR).

The sigma of variance-covariance matrix is `positive-semidefinite` and `symmetric` and the diagonal entries might be zero [1].
[1] https://en.wikipedia.org/wiki/Covariance_matrix#Properties

On the other hand, Cholesky decomposition requires  `positive-definite` and thus inverse of sigma cannot sometimes be resolved.
https://en.wikipedia.org/wiki/Cholesky_decomposition

Thoughts?

math/src/main/java/smile/math/matrix/CholeskyDecomposition.java
```diff
@@ -55,10 +55,17 @@
...
+     * @serial is symmetric and positive definite flag.
+     */
+    private final boolean isSPD;
+    
+    /**
      * Constructor. Used by newInstance().
      */
     private CholeskyDecomposition() {
-        
+        isSPD = true;
```
haifengl: Is this correct behavior? Note this constructor is used by newInstance(), which are used by QR and SVD.

haifengl:
Thanks! Can you please contribute your SDAR to smile?

myui:
@haifengl I'll close this ticket because I resolved my issue withou partial decomposition of CholeskyDecomposition, by using custom `pdf` function for Multivariate normal distribution.
Solver fallbacks to least square solution using SVD if matrix is singular.

PR #113, opened 2016-08-29
SOM Neuron Class Retrieval

KevinCooper:
Haifeng,
I am currently working with SOMs and your library.  While I can create a umatrix and plot it, I wanted to add a method to SOM to allow for the retrieval of class values for each of the Neurons.  Ideally, this would be run after calling partition(int).  This could be used to create more effective class graphs of the SOM.

### DIFF
--- a/core/src/main/java/smile/vq/SOM.java
+++ b/core/src/main/java/smile/vq/SOM.java
@@ -398,13 +398,13 @@ public class SOM implements Clustering<double[]> {
         if( y == null ){
             throw new IllegalStateException();
         }
-        int[][] classLabels = new int[height][width];
+        int[][] clusterLabels = new int[height][width];
         for (int i = 0, l = 0; i < height; i++) {
             for (int j = 0; j < width; j++) {
-                classLabels[i][j] = y[i*width + j];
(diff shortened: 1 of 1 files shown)


core/src/main/java/smile/vq/SOM.java
```diff
@@ -391,6 +391,19 @@ public SOM(double[][] data, int width, int height) {
     }
 
     /**
+     * Returns the cluster labels for each Neuron.
+     */
+    public int[][] clabels() {
```
haifengl: Can you please rename the function getClusterLabel, which is used by other clustering algorithms?

core/src/main/java/smile/vq/SOM.java
```diff
@@ -391,6 +391,19 @@ public SOM(double[][] data, int width, int height) {
...
+        int[][] clabels = new int[height][width];
```
haifengl: We should check if y is null at the beginning and throw an IllegalStateException if it is. Note y is only created after the method partition is called.

KevinCooper:
Haifeng,
The two proposed fixes have been made.  Thanks for your rapid response!

PR #127, opened 2016-11-15
making RandomForest serializable

alexeygrigorev:
Right now the inner class `RandomForest$Tree` is not serializable, so I get the following when I try to save the model:

```
Caused by: java.io.NotSerializableException: smile.classification.RandomForest$Tree
	at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1184)
	at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:348)
```

This simple modification should make it work


haifengl:
Thanks!

PR #131, opened 2016-12-06
ncol+attribute constructor for SparseDataset

alexeygrigorev:
Sometimes I want to specify both the number of columns and the type of the response variable when creating a sparse dataset. This small patch solves this problem.


haifengl:
Thanks!

PR #137, opened 2016-12-31
Simplify scala data package

jovilius:
Hi @haifengl, I've simplified a little bit the scala data package. What do you think?


haifengl:
Look good! Thank you very much!

PR #143, opened 2017-01-16
Update NeuralNetwork

lazymahendrathapa:
Update Neural Network to correct the bug in `setWeightDecay()` method.


haifengl:
Thanks!

PR #151, opened 2017-02-11
Correct correctOutputError() method when the error function is cross error.

lazymahendrathapa:
screenshot - 02112017 - 11 02 49 pm

The partial derivative of cost function with respect to z is (output - target) rather than (target - output) in the case of cross entropy.

### DIFF
--- a/core/src/main/java/smile/classification/NeuralNetwork.java
+++ b/core/src/main/java/smile/classification/NeuralNetwork.java
@@ -706,6 +706,7 @@ public class NeuralNetwork implements OnlineClassifier<double[]>, SoftClassifier
             } else if (errorFunction == ErrorFunction.CROSS_ENTROPY) {
                 if (activationFunction == ActivationFunction.SOFTMAX) {
                     error -= output[i] * log(out);
+                    g *= -1.0;
                 } else if (activationFunction == ActivationFunction.LOGISTIC_SIGMOID) {
                     error = -output[i] * log(out) - (1.0 - output[i]) * log(1.0 - out);
                 }


haifengl:
Have you run the test after your change?

sbt core/"test-only smile.classification.NeuralNetworkTest"

The error is much bigger. Not sure which book/paper you are referring about the gradient computing, I wonder if the update rules match too. Thanks!

haifengl:
A possible confusion is that for cross-entropy, the objective is to maximize it. While for mean-squared-errors, the objective is to minimize it. Our code has to handle them uniformly, so we have to revert the sign of cross-entropy.

lazymahendrathapa:
I am referring the slides of Geoffrey Hinton.

In case of cross-entropy, cost function is given by (-1.0 * summation(t*log(y))). How can our objective be to maximize it? y always range from 0-1. In the case of y less than zero, log(y) gives negative value and -1.0 * log(y) gives the positive value.

haifengl:
In this tutorial http://neuralnetworksanddeeplearning.com/chap3.html#introducing_the_cross-entropy_cost_function, the cross entropy cost function is a little different. And the gradient also has different sign.

Most importantly, you should run the unit tests. With your change, the error rate is way too high. Thanks!

lazymahendrathapa:
Thanks for being cooperative. I will study again about ANN from basic and try to understand everything mathematically and geometrically. SMILE is a great library, thanks for making open source. This helps lots of people including me. Can you send your contact mail at lazymahendrathapa@gmail.com, so that I can contact you if I stuck somewhere in future?

haifengl:
You can always ask me questions here. This place is better than email as everyone can see the discussions! Thanks!

PR #170, opened 2017-04-09
#fix: more correct and optimised version of range neighbor search

lifehackman:
#fix: more correct and optimised version of range neighbor search
The check on line 376: "radius >= diff * diff" - the diff is squared, but the radius isn`t. So when the radius is smaller than 1, the method is not optimal,
So I squared the radius internally and I squared the distance calculation, because it dramatically increases speed compared to calculating square root of the distance for every object. 
Sqrt is calculated only for suitable objects.

### DIFF
--- a/core/src/main/java/smile/neighbor/KDTree.java
+++ b/core/src/main/java/smile/neighbor/KDTree.java
@@ -356,7 +356,7 @@ public class KDTree <E> implements NearestNeighborSearch<double[], E>, KNNSearch
 
                 double distance = Math.distance(q, keys[index[idx]]);
                 if (distance <= radius) {
-                    neighbors.add(new Neighbor<double[], E>(keys[index[idx]], data[index[idx]], index[idx], distance));
+                    neighbors.add(new Neighbor<>(keys[index[idx]], data[index[idx]], index[idx], distance));
                 }
             }
         } else {
@@ -381,7 +381,7 @@ public class KDTree <E> implements NearestNeighborSearch<double[], E>, KNNSearch
 
     @Override
     public Neighbor<double[], E> nearest(double[] q) {
-        Neighbor<double[], E> neighbor = new Neighbor<double[], E>(null, null, 0, Double.MAX_VALUE);
+        Neighbor<double[], E> neighbor = new Neighbor<>(null, null, 0, Double.MAX_VALUE);
         search(q, root, neighbor);
         neighbor.distance = Math.sqrt(neighbor.distance);
         return neighbor;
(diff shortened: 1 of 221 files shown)


haifengl:
Thanks! Did you run unit tests on this change?

haifengl:
I manually pick your changes but I didn't change line 424 (instead I precompute radius * radius outside of for loop). I don't see any performance changes as I expected. A save of sqrt won't significantly change the speed. I guess that your speedup comes from the change at the line 424. Note that it is very dangerous that you pass radius * radius there but the "raw" radius is used in many places besides your changes at line 357. This change may cause wrong nearest neighbor search. I suggest that you run unit tests to test out your changes. Thanks!

core/src/main/java/smile/neighbor/KDTree.java
```diff
@@ -343,10 +343,10 @@ private void search(double[] q, Node node, HeapSelect<Neighbor<double[], E>> hea
      *
      * @param q    the query key.
      * @param node      the root of subtree.
-     * @param radius	the radius of search range from target.
+     * @param radiusSquared	the square of radius of search range from target.
```
haifengl: The point is not about to rename the radius.

core/src/main/java/smile/neighbor/KDTree.java
```diff
@@ -343,10 +343,10 @@ private void search(double[] q, Node node, HeapSelect<Neighbor<double[], E>> hea
...
      * @param neighbors the list of found neighbors in the range.
      */
-    private void search(double[] q, Node node, double radius, List<Neighbor<double[], E>> neighbors) {
+    private void search(double[] q, Node node, double radiusSquared, List<Neighbor<double[], E>> neighbors) {
         if (node.isLeaf()) {
             // look at all the instances in this leaf
```
haifengl: I add the following here:

double radius2 = radius * radius;

core/src/main/java/smile/neighbor/KDTree.java
```diff
@@ -355,7 +355,7 @@ private void search(double[] q, Node node, double radius, List<Neighbor<double[]
                 }
 
                 double distance = Math.squaredDistance(q, keys[index[idx]]);
-                if (distance <= radius) {
+                if (distance <= radiusSquared) {
```
haifengl: Here I compare it to radius2.

core/src/main/java/smile/neighbor/KDTree.java
```diff
@@ -370,11 +370,11 @@ private void search(double[] q, Node node, double radius, List<Neighbor<double[]
                 further = node.lower;
             }
 
-            search(q, nearer, radius, neighbors);
+            search(q, nearer, radiusSquared, neighbors);
```
haifengl: Don't change anything here and below.

lifehackman:
Sorry for misleading you. The sqrt really doesn`t impact performance. 
What really matters is this line of code:
```
double distance = Math.squaredDistance(q, keys[index[idx]]);
    if (distance <= radiusSquared) { ... }
...
if (radiusSquared >= diff * diff) {
    search(q, further, radiusSquared, neighbors);
}
```
compared to Nearest Neighbour search:

```
double distance = Math.squaredDistance(q, keys[index[idx]]);
if (distance < neighbor.distance) {
    ...
    neighbor.distance = distance;
}
...
// now look in further half
if (neighbor.distance >= diff * diff) {
     search(q, further, neighbor);
}
```
The diff (distance from cutting plane) is squared, and we must check it with square of distance, not with distance
I double checked the unit test, but it passed, as I understand (some time earlier I got assertion fails  because of not getting sqrt finally, and now i'm not getting them)
kdtreetest
Sorry for the lack of my github knowledge.

haifengl:
Can you please update the pull request? In general, I prefer that the changes are in local scope so that we don't get too many side effects. I am not exactly clear what you try to change here. Thanks!

lifehackman:
oh no, I have ruined everything. Maybe it is better to create a new branch and a new pull request, there is just one line of code that is changed

Issue #169, opened 2017-04-03
NominalAttribute#valueOf declares to throw ParseException


stieglma:
The method `NominalAttribute#valueOf(String s)` declares to throw `ParseException` which is not correct from my understanding. The method does only lookup strings in a map and if existant return the related double value. If not existant, a new relation between this string and an incremented number is added to the map.

Would be great if you could remove this unnecessarily declared exception. If you want to have a pull-request for that, I can also create one.

haifengl:
The declare of "throw ParseException" comes from the parent class Attribute. Note that what you describe is only true when the attribute is open. If not, we could throw an exception. I just update the code to throw ParseException instead of IllegalArgumentException. Thanks!

stieglma:
That's true, thanks for the fast fix! Do you already have a date in mind for the next release?

haifengl:
v1.3.0 was just released. So we will wait a little bit to cumulate enough fixes for next release. For now, you can clone and build the project by yourself. Thanks!

stieglma:
That's what I thought, we did clone the project already, and I also made some changes to clean up the code a little bit (many easy-to-fix rawtype/generics warnings coming up with eclipse, as well as some unused imports) if you are interested in that I can create a pull request with these fixes.

Exactly for the task of importing this project into eclipse was quite hard as there is no really good sbt-support. Would you mind creating a pom file such that it can be properly imported? This will make it for everyone who wants to contribute much easier to start.

haifengl:
Yes, please. Thank you very much!

We actually had pom files before. But since we switched to IntelliJ, the life is much easier with sbt. Then we removed pom files as they became outdated. It is hard to keep tracking two building system. More importantly, we have the scala module now.

haifengl:
v1.3.1 is released with all the bug fixes. Thanks!

PR #173, opened 2017-04-13
Code Cleanup

stieglma:
As promised in issue #169 here is the merge request with some code cleanup, involving fixing all kinds of rawtype warnings and removing some unused imports, there would be even more things to fix (unused variables ins tests, an unclosed resource in `ApproximateStringSearchDemo` and many missing `serialVersionUID` fields in serializable classes, but I am not sure if you intend this to be like that so I only fixed the first two mentioned things in this merge request.

### DIFF
--- a/core/src/main/java/smile/classification/RBFNetwork.java
+++ b/core/src/main/java/smile/classification/RBFNetwork.java
@@ -156,7 +156,7 @@ public class RBFNetwork<T> implements Classifier<T>, Serializable {
          * @param rbf the radial basis function.
          * @param m the number of basis functions.
          */
-        public Trainer setRBF(RadialBasisFunction rbf, int m) {
+        public Trainer<T> setRBF(RadialBasisFunction rbf, int m) {
             this.m = m;
             this.rbf = rep(rbf, m);
             return this;
@@ -166,7 +166,7 @@ public class RBFNetwork<T> implements Classifier<T>, Serializable {
          * Sets the radial basis functions.
          * @param rbf the radial basis functions.
          */
-        public Trainer setRBF(RadialBasisFunction[] rbf) {
+        public Trainer<T> setRBF(RadialBasisFunction[] rbf) {
             this.m = rbf.length;
             this.rbf = rbf;
             return this;
@@ -176,7 +176,7 @@ public class RBFNetwork<T> implements Classifier<T>, Serializable {
          * Sets true to learn normalized RBF network.
          * @param normalized true to learn normalized RBF network.
          */
-        public Trainer setNormalized(boolean normalized) {
+        public Trainer<T> setNormalized(boolean normalized) {
             this.normalized = normalized;
             return this;
         }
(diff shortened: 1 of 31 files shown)


core/src/main/java/smile/classification/RBFNetwork.java
```diff
@@ -156,7 +156,7 @@ public Trainer(Metric<T> distance) {
          * @param rbf the radial basis function.
          * @param m the number of basis functions.
          */
-        public Trainer setRBF(RadialBasisFunction rbf, int m) {
+        public Trainer<T> setRBF(RadialBasisFunction rbf, int m) {
```
haifengl: What's the warning about this? My IDE never report warnings about it. And previously another community member scanned the whole code base with lint tools. He didn't report something like this. I am just wonder why and is it because of Java 8? Thanks!

haifengl:
Thank you very much! I will merge the PR once I understand the warning.

core/src/main/java/smile/classification/RBFNetwork.java
stieglma: At first, no this isn't a Java 8 feature, this is in place since Generics are, I think this is since 1.5.

What happens there? Well Trainer itself is a class with a generic type parameter T (c.f. line 127 in the code, later on used for the distance field (`Metric<T>`)) and as such it should not refer to the "raw type" `Trainer` (without generic parameter) as you loose compile-time safety about the types used there.
Think of e.g. Lists, when you have a list of strings you write `List<String>` and get only offered methods to add strings to this list, and of course you can also only retrieve strings from that List (no ints, doubles, other types). But an even better example is the method `subList` which returns a List of Strings again, if they did omit the generic type parameter here, only a list would be returned where you do not know of which type the elements are, you would get only Objects instead everywhere. (This is actually what happens at runtime due to type erasure, but this is another topic ;) )

So in your specific case, the reason is that you wouldn't know which type is inside the trainer instance afterwards besides by using casts, which are not necessary if generics are used consistently.

stieglma: One further notice, in eclipse many of these warnings are generated directly when you write the code due to the incremental compiler, I am not sure how IntelliJ IDEA is working there but I have in mind that it doesn't compile instantly when you write code but only on certain actions. What I am quite sure of, is that IntelliJ does not by default run the so callled "code inspection" all the time, you have to trigger it manually, and it has a vast amount of configuration options. They are quite powerful and should definitely find and report such things.

stieglma: Sounds like a reasonable default, but I never heard of that, in fact you could also write something like this `public <X> Trainer<X> setRBF(...)` which could potentially make sense if you created a new trainer instance of different type there, so I think even if omitting the generic type parameter is valid you should add it such that this does not lead to problems in future development.

I also just tried that:

```
public class Test<T> {
        T tmp;

        public Test setTmp(final T x) {
            tmp = x;
            return this;
        }
        
        public T getTmp() {
            return tmp;
        }

        public Test<T> setTmp2(T y) {
            tmp = y;
            return this;
        }
    }
```

Now it is perfectly fine to write something like `String str = new Test<String>().setTmp2("abc").getTmp();`
but writing `String str = new Test<String>().setTmp("abc").getTmp();` does only work when you add a cast to `String`

haifengl:
Thanks!

PR #176, opened 2017-04-15
Fix: KDTree RNNSearch now does not search in redundant nodes

lifehackman:
if `abs(diff) < 1`, then  `diff * diff < abs(diff)`, so we are visiting redundant nodes, because we need only nodes where distance from cutting plane is less than radius

### DIFF
--- a/core/src/main/java/smile/neighbor/KDTree.java
+++ b/core/src/main/java/smile/neighbor/KDTree.java
@@ -373,7 +373,7 @@ public class KDTree <E> implements NearestNeighborSearch<double[], E>, KNNSearch
             search(q, nearer, radius, neighbors);
 
             // now look in further half
-            if (radius >= diff * diff) {
+            if (radius >= Math.abs(diff)) {
                 search(q, further, radius, neighbors);
             }
         }


core/src/main/java/smile/neighbor/KDTree.java
```diff
@@ -373,7 +373,7 @@ private void search(double[] q, Node node, double radius, List<Neighbor<double[]
             search(q, nearer, radius, neighbors);
 
             // now look in further half
-            if (radius >= diff * diff) {
+            if (radius >= Math.abs(diff)) {
```
haifengl: We has no assumption on `abs(diff) < 1` and I don't see anywhere we do that check. It may be true in your specific data. But we cannot simply assume it in general. Thanks!

lifehackman: but when the `abs(diff) > 1`, the original check might be wrong:
image

haifengl: I will cherry pick your change and test locally first. Thanks!

haifengl:
I add your changes to the master branch (I cannot merge yours directly as the line numbers don't match). I also add additional unit tests. Thanks!

The problem was probably because I compare distance to diff * diff in knn and nearest neighbor search, which are correct. I might just copy some logic from there. Glad you find it. Thanks!

lifehackman:
I myself use the highly optimized anti-library version with all basic types and inline convolution functions, because I need to range search for millions of objects, that's why I found this)
Thank you for maintaining the good ml library with rich possibilities!

haifengl:
Since you have millions of data, did you try LSH? It should be faster if you don't care it is an approximate algorithm.

BTW, what do you mean "anti-library version"? Have you compare smile to other nearest neighbor search libraries? Is smile fast enough? Thanks!

PR #178, opened 2017-05-02
Make methods in 'CooccurrenceKeywordExtractor' static

ariskk:
Since the `CooccurrenceKeywordExtractor` doesn't have any instance variables and the operations do not depend on an instance creation, I think it makes sense to make its methods static.
It makes sense in terms of class design and it should have performance benefit (in theory at least).


haifengl:
Although we have only one keyword extractor for now, I expect more keyword extraction algorithms in the future. Some of them will need setup (e.g. parameters). In general, we will have an interface KeywordExtractor and  all these algorithms will be implementations. So I would like to keep it non-static. A similar design can be found in tokenizer package. Thanks!

PR #195, opened 2017-06-21
Add symbolic classes and tests

defoye:
I must admit that I created this system more from intuition than from research.  There is a lot of work to be done to make it perfect, but I think it is off to an excellent start.

### DIFF
--- a/symbolic/src/main/java/smile/symbolic/Calculus.java
+++ b/symbolic/src/main/java/smile/symbolic/Calculus.java
@@ -1,0 +1,73 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Ernest DeFoy
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+
+package smile.symbolic;
+
+/**
+ * @author Ernest DeFoy
+ */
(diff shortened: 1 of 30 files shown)


haifengl:
Thanks! I merged it. The first thing we should do is to add more comments so that others can understand the code and contribute.

defoye:
Great.  I'll check out the rest of your project to get a feel for your style.  I can add "headers" like you have in some of your classes.  I can take care of all of them if you would like.

haifengl:
I made some changes to unit tests. Since all of our tests are based on jUnit 4. I change your test to it from jUnit 5 to avoid conflicts.

Besides, why do we have both CalcTest and CalculusTest? Can we combine them? Also the tests in CalculusTest fail.

defoye:
I started to rewrite that test class.  Yes, we can merge them.  The tests fail because I stopped developing the method for rewriting expressions.  I came up with a better way, a way in which we would have "rewrite rules," and check the expression tree for these rewrite patterns.  This sounds like a good plan to me.

defoye:
we can also use the rewrite rules to expand expressions, should they need to be for computation or expression rewriting

haifengl:
Cool. Let's add the headers to explain each class and its algorithm in high level. Then we can start the rewrite part. I have a lot of experience in DSL. Will look into your code and see what I can do. Honestly functional languages like ocaml and scala are much better choice for this task. But since the whole project is mainly based on Java, Java is still a good start. We can use scala to add a nicer interface. Thanks again!

defoye:
always glad to learn & contribute!

defoye:
Also, please be honest about the overall functionality of the code.  I am thick-skinned and can handle any criticism.  If you look at the system and decide that it doesn't fit with the rest of the project, I would be happy to change it, or even start over a different way(scala sounds interesting).  As someone with only a fraction of the experience that you have, I respect your input.

haifengl:
Sure thing

defoye:
Hi Haifeng, cool, sounds like a good plan.

Sure, I can remove the copyright, not a problem at all.

I am very busy and will only be busier come this fall, but I'm still here!

haifengl:
Thanks! If you can add more comments in the code, it would be very helpful.

PR #199, opened 2017-06-28
refine limit parameter value

redsnow1992:
`0` value will trail empty string and shorten splitted array, then cause `s.length != ncols`


haifengl:
I understand that you may face some issues when we trail empty string at the end for your file. But many .txt files have trail spaces without sense. This change will break the behavior. 

In general, we don't suggest to use space as an indication of missing values. A better choice is to use symbols like '?', which we support and you can customize.

If you have strings for that column, quote the string ("") is a good practice.

redsnow1992:
it is make sense

PR #203, opened 2017-07-24
SoftClassifierTrainer and StratifiedCrossValidation

noam-segev:
Added constructor with random number generator seed to cross-validation, allowing for development with constant seed.
Added SoftClassifierTrainer, an extension of ClassifierTrainer (where approriate and for whatever use).
Added StratifiedCrossValidation, making test/train sets share same class ration as the original data (for development/testing purposes)

### DIFF
--- a/core/src/main/java/smile/classification/SoftClassifierTrainer.java
+++ b/core/src/main/java/smile/classification/SoftClassifierTrainer.java
@@ -1,0 +1,38 @@
+package smile.classification;
+
+import smile.data.Attribute;
+
+/**
+ * Abstract soft classifier trainer.
+ * 
+ * @param <T> the type of input object.
+ * 
+ * @author Noam Segev
+ */
+public abstract class SoftClassifierTrainer <T> extends ClassifierTrainer<T> {
+	
+	/**
+     * Constructor.
+     */
+    public SoftClassifierTrainer() {
+        
+    }
+	
+	/**
+     * Constructor.
+     * @param attributes the attributes of independent variable.
+     */
+    public SoftClassifierTrainer(Attribute[] attributes) {
+		super(attributes);
+	}
+
+	/**
+     * Learns a soft classifier with given training data.
+     * 
+     * @param x the training instances.
+     * @param y the training labels.
+     * @return a trained classifier.
+     */
+    public abstract SoftClassifier<T> train(T[] x, int[] y);
+
+}
--- a/core/src/main/java/smile/validation/MissRate.java
+++ b/core/src/main/java/smile/validation/MissRate.java
@@ -1,0 +1,56 @@
+/*******************************************************************************
+ * Copyright (c) 2010 Haifeng Li
+ *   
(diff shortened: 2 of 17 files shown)


haifengl:
Thanks a lot for the contributions! I like your changes overall. Only two small suggestions:

1. The better way to set the seed is smile.math.Math.setSeed(). I have tried to avoid to use Random class directly in the code. But somehow missed the old usage in CrossValidation. I have fixed it and push the change to master. Please have a look. With this design, you will always get consistent results even with a random algorithm in a single threaded application (e.g. unit tests). In a multithreaded app, you should call Math.setSeed(), which is rarely needed in production.

2. In the future, please make a git pull request only for one feature. This request actually have 4 different things. It is faster to review and merge independent requests. Thanks again!

core/src/main/java/smile/classification/SoftClassifierTrainer.java
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	/**
+     * Learns a soft classifier with given training data.
+     * 
+     * @param x the training instances.
+     * @param y the training labels.
+     * @return a trained classifier.
+     */
+    public abstract SoftClassifier<T> train(T[] x, int[] y);
```
haifengl: @Override can be added here

core/src/main/java/smile/classification/SoftClassifierTrainer.java
```diff
@@ -0,0 +1,38 @@
+package smile.classification;
```
haifengl: Please add copyright header.

core/src/main/java/smile/validation/MissRate.java
```diff
@@ -0,0 +1,56 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+
+package smile.validation;
+
+/**
+ * Miss rate, or false negative rate (FNR)
```
haifengl: Do we really need this as FNR = 1 - sensitivity and we already have Sensitivity class?

core/src/main/java/smile/validation/StratifiedCrossValidation.java
```diff
@@ -0,0 +1,143 @@
...
+
+package smile.validation;
+
+/**
+ * An extension of Cross-validation for multi-class problems that adds an
+ * assertion that all partition keep a similar ratio between classes as the
+ * original data.
+ *
+ * @author Noam Segev
+ */
```
haifengl: Shall we add this to CrossValidation class with an optional parameter "boolean stratified" to control the behavior? The default value should be false so that we won't break existing code.

haifengl:
There is no need SoftClassifierTrainer as all ClassifierTrainer specializes the return type. BTW, we have more than two types of Classifier. Your design won't work for multiple types.

PR #211, opened 2017-08-20
working on scala API for classification

anuj1207:
*updated some methods with functional style
*tried to remove null
*fixed some typos for camel casing

### DIFF
--- a/benchmark/src/main/scala/smile/benchmark/Airline.scala
+++ b/benchmark/src/main/scala/smile/benchmark/Airline.scala
@@ -101,7 +101,7 @@ object Airline {
     // AdaBoost
     test2soft(x, y, testx, testy) { (x, y) =>
       println("Training AdaBoost of 300 trees...")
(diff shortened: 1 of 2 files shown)


haifengl:
Thanks for your effort! However, we don't use camel case here on purpose. We choose these names to be compatible (or similar) with R or Matlab. The usage of null for parameter is to be same as the Java API.

In the future, please communicate first for styling things. There is no right or wrong on styling. But there are reasons why I choose the current style. Thanks!

anuj1207:
Thanks, sir! I understand the camel casing but I was hoping you would use the functional style for scala API. And as for null, I was hoping you would want to work on creating Scala classes for Decision Tree and others in future.
Thank you

haifengl:
Do you mean implement decision tree from scratch with Scala? If so, it is not necessary. 

BTW, I don't really see a big difference between for (i <- 0 until 100) and (0 until 100) foreach. These small syntax difference won't cause the difference on code readability and performance.

PR #215, opened 2017-08-30
Explicit SBT version

pfcoperez:
Specify SBT version as a project property. 

That way, local updates won't break the compilation.


haifengl:
Thanks!

PR #220, opened 2017-09-01
math: GaussianKernel can auto-compute gamma

mgiannini:
This patch adds the ability to configure the Gaussian Kernel to auto-compute gamma as `1.0 / num_features`. This is a nice convenience to have and adds feature parity with other machine learning libraries.


haifengl:
I prefer that an explicit configuration. It is not really bad just passing in 1/x.length in constructor.

mgiannini:
That makes sense. Thanks for the quick feedback.

PR #231, opened 2017-09-20
Add Mutual Information and Normalized Mutual Information

beckgael:
Hi, 
Everything is on the commit name.
Results are similar to scikit learn library 
http://scikit-learn.org/stable/modules/generated/sklearn.metrics.normalized_mutual_info_score.html


scala/src/main/scala/smile/validation/Normalize_Mutual_Index.scala
```diff
@@ -0,0 +1,67 @@
+import scala.math.{max, log, sqrt}
+
+/**
+ * @author Beck Gaël
+ */
+object Externals_Indexes
```
haifengl: In general, the file name should not contain underscore _ and the class name is better same as the file name. It is just a matter of style. Better be consistent with the rest of library.

More importantly, why is this in smile.validation package? For fundamental math functions like this, I suggest too add it to smile.math.Math. Also the Java users can leverage it too.

scala/src/main/scala/smile/validation/Normalize_Mutual_Index.scala
```diff
@@ -0,0 +1,67 @@
+import scala.math.{max, log, sqrt}
```
haifengl: Can you copy our copyright header here? Thanks!

scala/src/main/scala/smile/validation/Normalize_Mutual_Index.scala
```diff
@@ -0,0 +1,67 @@
...
+	 * Normalize Sequences in order to prevent construction of a to big 'count' matrix
+	 * Ex: [4,5,6,6] -> [0,1,2,2]
+	 **/
+	private def prepareList(x: Array[Int]) =
+	{
+		val indexedValuesMap = x.distinct.zipWithIndex.toMap
+		x.map(indexedValuesMap)
+	}
+
+	private def mutualInformationInteral(x: Array[Int], y:Array[Int]) =
```
haifengl: Comments please. Why integer array? Are they counts or index of binary sparse array?

scala/src/main/scala/smile/validation/Normalize_Mutual_Index.scala
```diff
@@ -0,0 +1,67 @@
...
+	{
+		require( x.size == y.size )
```
haifengl: The implementation is overall imperative rather than functional. So why not we implement it in Java so that Java users can use it too? Again, for fundamental things, I prefer Java for bigger audience. For high algorithms, I am fine with Scala implementation as long as the interface can be called from Java too.

Sorry for miscommunication in earlier gitter chat.

scala/src/main/scala/smile/validation/Normalize_Mutual_Index.scala
```diff
@@ -0,0 +1,67 @@
...
+		bj.foreach( v => { val c = v / nN; if( c > 0) hv -= c * log(c) } ) 
+
+		var huStrichV = 0D
+		for( i <- maxOneIndices ) for( j <- maxTwoIndices ) if( count(i)(j) > 0 ) huStrichV -= count(i)(j) / nN * log( (count(i)(j)) / bj(j) )
+
+		val mi = hu - huStrichV
+		(mi, hu, hv)
+	}
+
+	def mutualInformation(x: Array[Int], y:Array[Int]) = mutualInformationInteral(x, y)._1
```
haifengl: Overall, please add enough comments.

scala/src/main/scala/smile/validation/Normalize_Mutual_Index.scala
```diff
@@ -0,0 +1,67 @@
...
+
+	def nmi(x: Array[Int], y: Array[Int], normalization: String = "sqrt") =
```
haifengl: can we a little bit consistent with the naming? nmi is not clear for a starter and not consistent with the naming convention of mutualInformation().

It is understandable to use a string in Python for the parameter "normalization". But for a strong type language, String is a bad idea. Even worse, without documentation, who knows what should I pass in for this parameter.

haifengl:
Reading the python link. So this is not a generic mutual information function but a mutual information score for validation?

haifengl:
Since this is a metric to evaluate clustering, please change it from `object` to `class` and implements the interface `ClusterMeasure`. Scala is fine. Please ignore some of my previous comments. `RandIndex` is an example to follow. Instead of providing two functions, move `normalization` parameter to the constructor.

The reason that it is a class instead of single instance object is that we can pass a set of `ClusterMeasure` as generic measures. Thanks!

scala/src/main/scala/smile/validation/ClusterMeasure.scala
```diff
@@ -0,0 +1,35 @@
+
```
haifengl: You don't need to define `ClusterMeasure`. It is already defined in Java and Scala can implement a Java interface.

scala/src/main/scala/smile/validation/MutualIndex.scala
```diff
@@ -0,0 +1,28 @@
+/*******************************************************************************
```
haifengl: Three class design is over-complicated. A single class is enough. The default `normalization` parameter in constructor can be `None`.

scala/src/main/scala/smile/validation/NormalizedMutualIndex.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import scala.math.{max, log, sqrt}
+
+/**
+ * @author Beck Gaël
+ */
+
+ /**
+  * Computte the normalized mutual information
+  **/
```
haifengl: Please list all valid values of `normalization` in document.

scala/src/main/scala/smile/validation/NormalizedMutualIndex.scala
```diff
@@ -0,0 +1,40 @@
...
+class NormalizedMutualInformation(normalization: String = "sqrt") extends MutualInformationInternal
+{
+	override def measure(x: Array[Int], y: Array[Int]) =
+	{
+		val (mi, hu, hv) = mutualInformationInternal(x, y)
+		val nmi = normalization match
+		{
+			case "sqrt" => mi / sqrt(hu * hv)
+			case "max" => mi / max(hu, hv)
+			case _ => mi / sqrt(hu * hv)
```
haifengl: `case _` for String is bad idea in general. Exception should be thrown for invalid values.

scala/src/main/scala/smile/validation/NormalizedMutualIndex.scala
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * @author Beck Gaël
+ */
+
+ /**
+  * Computte the normalized mutual information
+  **/
+class NormalizedMutualInformation(normalization: String = "sqrt") extends MutualInformationInternal
+{
```
haifengl: please check our scala code for styling. We don't put `{` in the next line.

scala/src/main/scala/smile/validation/MutualIndexInternals.scala
```diff
@@ -0,0 +1,76 @@
...
+package smile.validation
+
+import scala.math.{max, log, sqrt}
+
+/**
+ * @author Beck Gaël
+ */
+
+ /**
+  * Computte the mutual information
```
haifengl: Detailed information about the measure should be given. What is its meaning, the formula, what do they try to detect, etc.

haifengl:
Thanks! I merged the pull request and also made some minor revisions. Can you please also implement the adjust mutual information score?

PR #237, opened 2017-10-15
 Copy epochs attribute in clone method for NeuralNetwork 

serickso:


haifengl:
Hi,

After merge the pull request, I decided to remove the epochs member variable. It can be part of the Trainer but should not be part of the network itself. Note that the learn() should go through the samples only once. The caller has the flexibility how many epochs to train the model. Thanks!

PR #239, opened 2017-10-18
Added dot methods for matrix multiplication in Math.java

serickso:


haifengl:
There were matrix multiplication methods in Math.java but I removed them on purpose. They are very slow because of the design of Java. It is recommended to use smile.math.matrix.Matrix for this operation.

PR #240, opened 2017-10-19
 Implemented OnlineRegression for OLS 

serickso:


core/src/main/java/smile/regression/OLS.java
```diff
@@ -22,6 +22,7 @@
 import org.slf4j.LoggerFactory;
 import smile.math.Math;
 import smile.math.matrix.Matrix;
+import smile.math.matrix.JMatrix;
```
haifengl: JMatrix should never be directly used. Instead, you should declare a matrix as DenseMatrix and use Matrix.newInstance() to create one. In this way, the system will automatically pick up a native implementation if it is available. If a native implementation is not available, it will fall back to JMatrix.

core/src/main/java/smile/regression/OLS.java
```diff
@@ -96,6 +97,10 @@
      */
     private double[][] coefficients;
     /**
+     * The gamma matrix for recursive least squares used in online updates
```
haifengl: What exactly is the gamma matrix? Can you add comments how gamma is defined and used?

core/src/main/java/smile/regression/OLS.java
```diff
@@ -253,10 +259,18 @@ public OLS(double[][] x, double[] y, boolean SVD) {
                     double t = w1[i] / se;
                     coefficients[i][2] = t;
                     coefficients[i][3] = Beta.regularizedIncompleteBetaFunction(0.5 * df, 0.5, df / (df + t * t));
+                    if (!Math.isZero(s,1E-5)){
```
haifengl: Why different value for gamma(i,i)?

core/src/main/java/smile/regression/OLS.java
```diff
@@ -253,10 +259,18 @@ public OLS(double[][] x, double[] y, boolean SVD) {
...
+                        gamma.set(i,i,1/(s*s));
+                    }
+                    else {
+                        gamma.set(i,i,1/s);
```
haifengl: If s is close to zero, this will be very large. It is not numerically stable.

core/src/main/java/smile/regression/OLS.java
```diff
@@ -376,6 +391,54 @@ public double predict(double[] x) {
...
+    
+    /**
+     * Learn a new instance with online regression.
+     * @param x the training instance.
+     * @param y the target value.
+     */
+    @Override
+    public void learn(double[] x, double y){
+        JMatrix X = new JMatrix(p+1,1);
+        JMatrix W = new JMatrix(p+1,1);
```
haifengl: It is not efficient to create W for each sample. Similarly for matrix X.

core/src/main/java/smile/regression/OLS.java
```diff
@@ -376,6 +391,54 @@ public double predict(double[] x) {
...
+        for (int i=0; i<p; i++){
```
haifengl: System.arraycopy() is more efficient.

core/src/main/java/smile/regression/OLS.java
```diff
@@ -376,6 +391,54 @@ public double predict(double[] x) {
...
+            W.set(i,0,w[i]);
+        }
+        X.set(p,0,1);
+        W.set(p,0,b);
+        updateGamma(X);
+        updateW(X,W,y);
+    }
+    
+    private void updateGamma(JMatrix x){
+        double v = 1+gamma.xax(x.data());
```
haifengl: please add space between operator and parameters. Please keep consistent with the style of existing code. For example, we would like to format it as

```
private void updateGamma(JMatrix x) {
        double v = 1 + gamma.xax(x.data());
```

core/src/main/java/smile/regression/OLS.java
```diff
@@ -75,7 +76,7 @@
  * 
  * @author Haifeng Li
  */
-public class OLS implements Regression<double[]>, Serializable {
+public class OLS implements Regression<double[]>, OnlineRegression<double[]>, Serializable {
```
haifengl: OnlineRegression is a subclass of Regression. We don't both of them here.

serickso: I did this to support better documentation. Using both Regression and OnlineRegression interfaces allows users to identify classes that implement OnlineRegression in the documentation for OnlineRegression, the "All Known Implementing Classes:" section. The same idea holds for the Regression interface. 

What do you think?

core/src/main/java/smile/regression/OLS.java
serickso: Gamma is the covariance matrix (X'X)^-1. You can find out more about this at https://en.wikipedia.org/wiki/Online_machine_learning#Online_learning:_recursive_least_squares

I decided to used only the diagonal elements for gamma for computational reasons (for SVD we would have to compute the inverse of VSS'V', since X'X=VSS'V'. This can be done using the pseudoinverse) and by some tests I did where the MSE for using the full covariance matrix were very close to using only the singular values from S or the diagonal elments from the cholesky inverse.

I can change gamma to be the full covariance matrix, if you want me to.

core/src/main/java/smile/regression/OLS.java
```diff
@@ -252,17 +285,26 @@ public OLS(double[][] x, double[] y, boolean SVD) {
...
                     double t = w1[i] / se;
                     coefficients[i][2] = t;
-                    coefficients[i][3] = Beta.regularizedIncompleteBetaFunction(0.5 * df, 0.5, df / (df + t * t));
+                    coefficients[i][3] = Beta.regularizedIncompleteBetaFunction(0.5 * df, 0.5, df / (df + t * t)); 
                 } else {
                     coefficients[i][1] = Double.NaN;
                     coefficients[i][2] = 0.0;
                     coefficients[i][3] = 1.0;
                 }
+                if (!Math.isZero(s, .01)){
```
haifengl: Why 0.01? It seems very large to me. Do you have a reference on this choice?

core/src/main/java/smile/regression/OLS.java
```diff
@@ -252,17 +285,26 @@ public OLS(double[][] x, double[] y, boolean SVD) {
...
+                    sigmaInverse.set(i, i, 1/Math.sqr(s));
+                }else{
+                    sigmaInverse.set(i, i, 1);
```
haifengl: Why 1? 0 seems right to me for inverse matrix.

core/src/main/java/smile/regression/OLS.java
```diff
@@ -252,17 +285,26 @@ public OLS(double[][] x, double[] y, boolean SVD) {
...
+                }
             }
+            // Calculate pseudoinverse of X^T * X
+            DenseMatrix V = svd.getV();
+            onlineParameters.gamma = V.abmm(sigmaInverse.abmm(V.transpose()));
```
haifengl: don't do transpose V.transpose(). just call sigmaInverse.abtmm(V). It is more efficient.

core/src/main/java/smile/regression/OLS.java
```diff
@@ -376,6 +418,58 @@ public double predict(double[] x) {
...
+        b -= delta[p];
+        onlineParameters.W.set(p, 0, b);
+    }
+    
+    /**
+     * Update the weights with given instance and associated target value.
+     * @param x the training instances.
+     * @param y the target values.
+     */
+    public void learn(double[][] x, double[] y){
```
haifengl: People may be confused with this method and the learning in constructor. Do we have to add this method?

core/src/main/java/smile/regression/OLS.java
```diff
@@ -75,10 +75,36 @@
...
+        * The gamma matrix is updated with each learning instance in the learn()
+        * method, and it is used to update the coefficients in online learning.
+        * Gamma is first initialized to the matrix (X<sup>T</sup>X)<sup>-1</sup>.
+        * 
+        * A more complete description can be found on the wikipedia article
+        * about online machine learning.
+        */
+        DenseMatrix gamma;
+       /**
+        * A single learning instance X
```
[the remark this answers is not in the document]
serickso: I did it this way to get gamma*X*X^T*gamma, for the X*X^T part. A more efficient way could be:
double[] gammax=gamma.ax(x)
double[] xtGamma = gamma.atx(x)
// compute gammax * xtGamma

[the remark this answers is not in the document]
haifengl: Matrix-vector multiplication is much better than matrix-matrix multiplication in this case. In fact, you just need gamma.ax(x) and then calculate the norm. No need gamma.atx(x).

core/src/main/java/smile/regression/OLS.java
serickso: I used 0.01 because I was worried about 1/s^2 becoming very large, and potentially causing the elments in the gamma matrix becoming large and unstable. 

If you ignore the fact that I don't place 1/s^2 in sigmaInverse in cases where s is in (-0.01, 0.01), then it can be shown that the gamma matrix after the ith step is equal to (X^T * X)^-1, where X contains all initial input observations, and all input observations up to and including the ith step. 

So, the online learning solution after i learning steps is equivalent to finding coefficients with a traditional OLS solution. Maybe the best way to handle small values when initializing the gama matrix is to deal with them as you originally do in solve() methods in SVD and QR.

serickso:
I think this will be easier to use if we place it in a separate class entirely, called RLS, short for recursive least squares. This has the following benefits:
1. Users who actually need to on online model for linear regression won't get confused.
2. Users who only want to use OLS won't get confused.
3. Having separate classes will make maintenance easier for developers.

core/src/main/java/smile/regression/OLS.java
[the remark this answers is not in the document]
haifengl: Gamma X X ^T Gamma = (Gamma X)(Gamma X) ^T because Gamma is symmetric.

haifengl:
It maybe a good idea to separate online learning version out. Most time, people use the batch version and it is not necessary to calculate all extra work for online learning part.

serickso:
Ok, I have added the RLS and RLSTest.

haifengl:
Thank you!

haifengl:
I clean up the code a little bit.

1. Always use SVD instead of QR because it is more stable when the data is close to rank deficient, which is more likely in RLS as the initial data size may be small.
2. remove duplicate w, b, and W
3. keep only the constructor with batch initialization. The other don't initialize gamma/V properly.

PR #242, opened 2017-10-30
Kept exception messages consistent, recalculated Vx before updating w

serickso:


haifengl:
why do we need updatedVx? It is same as Vx, right?

serickso:
It changes because of the update to V. First we update V, then use the updated V to update w

core/src/main/java/smile/regression/RLS.java
```diff
@@ -209,12 +214,17 @@ public void learn(double[] x, double y) {
             for (int i = 0; i <= p; i++) {
                 double tmp = V.get(i, j) - ((Vx[i] * Vx[j])/v);
                 V.set(i, j, tmp/lambda);
+                
+                if (j == 0){
+                    updatedVx[i] = 0;
+                }
+                updatedVx[i] += V.get(i, j) * x1[j];
             }
         }
```
haifengl: just add 
```
V.ax(x1, Vx)
```
here. And remove updatedVx.

haifengl:
don't do matrix operation by yourself. ax() is much more efficient.

haifengl:
Thank you!

PR #246, opened 2017-11-03
Implemented RNN for regression

serickso:
I haven't worked out any potential improvements to the performance yet.


serickso:
Instead of worrying about the whole memory thing I changed in my last commit, I think it might just be easier to use the batch mode for the RNN instead. Then implement a learn method that takes double[][] x and double[] y as input, and reset the h_{t-1} variable to 0s in h_t = sigma(Wxh x Whh h_{t-1}) every time learn is called. 

My reasoning is that this will make time series forecasting easier for users, so that if a user learns some sequence of points (X1, y1), (X2, y2), . . ., (Xn, yn), then they can predict Xn+1 through Xn+k. It will help

serickso:
I think it is better to convert RNN as an abstract base class. This would ensure proper behavior in RNNs that don't have issues with vanishing or exploding gradients, such as LSTM or GRU. To get started with a more comman RNN right away, I could just convert the existing RNN to an abstract base class and implement an Elman network.

haifengl:
Yes, it is nice to make it a base class.

PR #251, opened 2017-11-23
Ignoring some columns with DelimitedTextParser

gcoter:
Adds the possibility to ignore some columns with DelimitedTextParser. By default, all columns are read. You can add column indices to ignore with the methods `setIgnoredColumns`, `addIgnoredColumn` and `addIgnoredColumns`. If `hasRowNames` is true, the first column is ignored. The response variable cannot be ignored.


haifengl:
Thanks!

PR #254, opened 2017-11-28
Negative Values in Bernoulli Constructor should throw correct error 

MartinSchmitzDo:
fixed a bug which caused an ArrayOutOfBoundException in case of negative integers in data[]


haifengl:
Thanks!

PR #255, opened 2017-11-30
Add additional constructor to CrossValidation

serickso:
Add additional constructor to CrossValidation so users can choose if they don't want the indices to be permutated.


haifengl:
thanks!

PR #268, opened 2018-02-08
Add support for attribute weights in regression Random Forest

Ajk4:
I just multiply split score by attribute weight

### DIFF
--- a/core/src/main/java/smile/regression/RegressionTree.java
+++ b/core/src/main/java/smile/regression/RegressionTree.java
@@ -585,6 +585,9 @@ public class RegressionTree implements Regression<double[]>, Serializable {
                 throw new IllegalStateException("Unsupported attribute type: " + attributes[j].getType());
             }
 
+            double attributeWeight = attributes[j].getWeight();
+            split.splitScore *= attributeWeight;
+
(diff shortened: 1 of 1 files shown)


haifengl:
Any theory support for this feature? Paper reference?

Ajk4:
Hi  @haifengl  

I'm not exactly sure what you meant.

Do you want to check paper references regarding specific implementation details? 

Or do you want to check paper references whether it make sense to weight features in the first place? If the latter maybe I can drop some links to R implementation that have this feature.

RBorist:

https://cran.r-project.org/web/packages/Rborist/Rborist.pdf

Example usage:
```
# Causes first three predictors to be selected as splitting candidates
# twice as often as the other two:
rb <- Rborist(x, y, predWeight=c(2.0, 2.0, 2.0, 1.0, 1.0))
```

Another alternative could be providing some optional function parameter so user can provide custom split scoring function. That way if you don't like having weighted features here user could provide in it's own scoring function.

What do you think?

haifengl:
We need theoretical support that this is a legitimate operation. I don't feel that this operation makes too much sense. What Rborist does is not important.

PR #269, opened 2018-02-08
Add regression random forest merging capability

Ajk4:
Add an option to merge two random forests into one

Example use cases:
* Build random forests on spark executors and then merge them together into one big forest on driver.
* Grow forest iteratively over time

### DIFF
--- a/core/src/main/java/smile/regression/RandomForest.java
+++ b/core/src/main/java/smile/regression/RandomForest.java
@@ -489,14 +489,43 @@ public class RandomForest implements Regression<double[]>, Serializable {
         if (m > 0) {
             error = Math.sqrt(error / m);
         }
-        
-        importance = new double[attributes.length];
+
+        importance = calculateImportance(trees, attributes.length);
+    }
+
+    /**
+     * Merges together two random forests and returns a new forest consisting of trees from both input forests.
+     */
+    public RandomForest merge(RandomForest other) {
+        if (this.importance.length != other.importance.length) {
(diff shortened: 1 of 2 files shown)


Ajk4:
Also note, that data needed for proper merge error estimation is already lost, so I am doing an weighted average (in respect to forest size) of errors from both forest

serickso:
@Ajk4 could you provide some references for your pull request?

Ajk4:
Hi @serickso 

I doubt that there would be a paper reference since it's a technical detail. Could you try to be more specific about what you expect?

Random forest is List of DecisionTree's. Merging two Lists of decision trees into one is something that might be (and in my case it is) very useful.

With current implementation it's not possible, since there are no constructor accepting list of decision trees.

If you don't like adding new 'merge' method what about adding public constructor accepting list of decision trees? 

That way in users code we could do something like
```
RandomForest rf1 = (...); // from some spark executor
RandomForest rf2 = (...); // from other spark executor

List<DecisionTree> allTrees = Lists.concat(rf1.getTrees(), rf2.getTrees())
RandomForest combined = new RandomForest(allTrees);
```

Now that I see it I like it more that adding `merge` function.

Ajk4:
Some reference, in R there is `combine` function for that -> https://github.com/cran/randomForest/blob/master/man/combine.Rd

haifengl:
I will add a combine function in scala package. This should not be in the core algorithm. Thanks.

haifengl:
Thanks!

PR #272, opened 2018-02-12
Working around RNG issues with UniversalGenerator

Ajk4:
### TL:DR 

If you are unlucky, your random forest will use at most ~10 datapoints from train set.

### Diving into the source code

I was debugging random forest that would not fit my train data.

Using break-points I noticed something very suspicious. 

https://github.com/haifengl/smile/blob/554b85cef4b7520c3483a2e4739bebaf9559e398/core/src/main/java/smile/regression/RandomForest.java#L292

After this loop `samples` array had mostly zeros, and only some indices had unusual high values. It was clear that random index was not so random.

I noticed, that the only random numbers for incides I got were: {63,64,127,128}. It looks like some kind of bit-operation issues.

Now, after debugging I have simple repro. Just run this code:

```
public class UniversalGeneratorBug {

    public static void main(String[] args) {
        UniversalGenerator random = new UniversalGenerator();

        HashSet<Integer> uniqueNumbers = new HashSet<>();
        for (int i = 0; i < 100000; i++) {
            uniqueNumbers.add(random.nextInt(320));
        }

        for (Integer num : uniqueNumbers) {
            System.out.println(num);
        }
    }

}
```
It will print:
```
128
256
192
0
64
319
255
191
127
63
```

Only 10 unique numbers from range 0-320! Certainly not random.

#### Some more observations

I noticed that nextInt() results will always have all least significant bits set all to 0 or all to 1. 

In my case 320 = 256 + 64. 

I guess this introduces some symmetry that results in regular outputs from RNG.

#### Workaround

I switched default implementation to MersenneTwister. I checked that Mersenne Twister does not suffer from this issue.

Also fixed some infinite recursion that will result in StackOverflow when called

### DIFF
--- a/math/src/main/java/smile/math/Random.java
+++ b/math/src/main/java/smile/math/Random.java
@@ -1,10 +1,10 @@
 /*******************************************************************************
  * Copyright (c) 2010 Haifeng Li
- *   
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *  
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
@@ -16,13 +16,13 @@
 
 package smile.math;
 
+import smile.math.random.MersenneTwister;
 import smile.math.random.RandomNumberGenerator;
-import smile.math.random.UniversalGenerator;
 
 /**
  * This is a high quality random number generator as a replacement of
  * the standard Random class of Java system.
- * 
+ *
  * @author Haifeng Li
  */
 public class Random {
@@ -33,14 +33,14 @@ public class Random {
      * Initialize with default random number generator engine.
      */
     public Random() {
-        this(new UniversalGenerator());
+        this(new MersenneTwister());
     }
 
     /**
      * Initialize with given seed for default random number generator engine.
      */
     public Random(long seed) {
-        this(new UniversalGenerator(seed));
+        this(new MersenneTwister(seed));
     }
 
     /**
@@ -52,6 +52,7 @@ public class Random {
 
     /**
      * Generator a random number uniformly distributed in [0, 1).
+     *
      * @return a pseudo random number
      */
     public double nextDouble() {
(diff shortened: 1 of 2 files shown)


haifengl:
This is very interesting. I will try to figure out what's wrong ASAP. Thanks a lot!

haifengl:
I fixed the nextInt of UniversalGenerator. Thanks.

Ajk4:
Hi @haifengl

    Underlying nextDouble() function is still probably bugged. Least significant bits in it are always zeros. Perhaps original algorithm was not designed to work with 64 double floating numbers, but with less bit numbers instead?

I feel like such workaround will bite us again eventually. Could you consider merging PR I added that switches the implementation to MersenneTwister?

    Additionally we could consider adding java.random implementation of smile random interface just so users can switch to it if needed.

haifengl:
UniversalGenerator and MersenneTwister work in different ways. MersenneTwister generates random bits (and thus integer) and then we scale it to double value in [0, 1). On the other hand, UniversalGenerator generates real values directly in [0, 1) then we scale it to an integer. Since it generates values only in [0, 1), not all bits would be used by nature. The bug was caused by that we adopt MersenneTwister's nextInt to UniversalGenerator. Because they are not compatible algorithms, this cause the issues.

Either way has its pros and cons. It is probably a good idea that only use UniversalGenerator for nextDouble while MersenneTwister for nextInt.

BTW, you pull requests includes a lot of unnecessary white space changes, which is not desired. Thanks.

Ajk4:
Okay! Thanks for clarification. TIL about some of those RNG algorithms.

> Since it generates values only in [0, 1), not all bits would be used by nature.

That is true. 

I should have clarified my comment - fraction part of double number has all zeros on least significant bits, while those bits could be randomized while it's still [0, 1) range.

Anyways, this probably doesn't matter now that we have your fix for Integer values and we probably don't need randomness that far after decimal point.

> BTW, you pull requests includes a lot of unnecessary white space changes, which is not desired.

Okay - I'll keep this in mind for future commits to not introduce unrelated white space changes.

Those white spaces were automatically applied by my IDE (InteliJ). It's probably cleaner not to have orphan spaces in random places so we might have one commit that automatically cleans up all those orphan spaces :P 

@haifengl 

I extracted infinite-recursion fix to another PR -> https://github.com/haifengl/smile/pull/274

haifengl:
In smile.math.Random, we use UniversalGenerator for nextDouble and MersennaTwister for nextInt now.

PR #274, opened 2018-02-13
Fix infinite recursion when calling setSeed with long

Ajk4:
Calling setSeed on MersenneTwister used to result in StackOverflow with infinite recursion

### DIFF
--- a/math/src/main/java/smile/math/random/MersenneTwister.java
+++ b/math/src/main/java/smile/math/random/MersenneTwister.java
@@ -75,7 +75,7 @@ public class MersenneTwister implements RandomNumberGenerator {
 
     @Override
     public void setSeed(long seed) {
-        setSeed((int)seed % UniversalGenerator.BIG_PRIME);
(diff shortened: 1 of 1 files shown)


math/src/main/java/smile/math/random/MersenneTwister.java
```diff
@@ -75,7 +75,7 @@ public MersenneTwister(int seed) {
 
     @Override
     public void setSeed(long seed) {
-        setSeed(seed % UniversalGenerator.BIG_PRIME);
+        setSeed((int)seed % UniversalGenerator.BIG_PRIME);
```
haifengl: It should be
```
setSeed((int) (seed % UniversalGenerator.BIG_PRIME));
```
Otherwise, we will lose useful digits in the `seed`. Thanks.

Ajk4: Good catch. I fixed it in new commit

PR #281, opened 2018-03-02
Add monotonic regression capabilities to RegressionTree

Ajk4:
Hi, 

this PR adds support for monotonic regression in regression trees. For more details you can check docs from XGBoost that have this feature -> http://xgboost.readthedocs.io/en/latest/tutorials/monotonic.html

For some references you can check those slides 
http://past.rinfinance.com/agenda/2016/talk/MarkSeligman.pdf 
or already linked XGBoost doc at http://xgboost.readthedocs.io/en/latest/tutorials/monotonic.html .

BTW While working on it I created some auxiliary commits with some refactors and cleaning. 

For example I have separate commit that fixes formatting so you have easier review and it doesn't introduce noise to commit that implements actual feature.

If you would like to have MonoReg merged but without some of those auxiliary commits please let me know - I could probably extract just the MonoReg feature here.


haifengl:
Thanks. I am refactoring DecisionTree/RegressionTree for performance improvement. After that, I may merge this feature.

data/src/main/java/smile/data/Attribute.java
```diff
@@ -63,6 +63,16 @@
      */
     private double weight;
 
+    /**
+     * Value between [-1, 1] that represents monotonic regression coefficient.
```
haifengl: Is it necessary to add this variable? It is algorithm dependent if a attribute is used for monotonic regression. It is sufficient to define this in the algorithm API, rather than here.

Ajk4: Yeah, I see what you mean. I think I will change it that way. 

This Attribute is used for things other than regression trees too. And monotonic regression makes sense only for regression tree model. So it's probably better to have it in Algorithm API

Ajk4:
Hi @haifengl 

I pushed commit with your suggestion and moved monoreg param to the regression tree constructor.

Could you maybe review my commits and accept the changes if possible? Thanks!

data/src/main/java/smile/data/Attribute.java
```diff
@@ -73,30 +73,18 @@
      */
     private String description;
```
haifengl: Please reverts the changes to Attribute.java and NumericAttribute.java

Ajk4: Done

haifengl:
Thanks!

PR #286, opened 2018-03-09
Fixes FP-Growth single path FP-tree mining

peter-toth:
Hi,
This PR fixes the issue in FPGrowth when mining a single path tree.
Cheers,
Peter

### DIFF
--- a/core/src/main/java/smile/association/FPGrowth.java
+++ b/core/src/main/java/smile/association/FPGrowth.java
@@ -317,6 +317,42 @@ public class FPGrowth {
     }
 
     /**
+     * Mines all combinations along a single path tree
+     */
+    private long grow(PrintStream out, List<ItemSet> list, TotalSupportTree ttree, FPTree.Node node, int[] itemset, int support) {
+        int n = 0;
+        node = node.parent;
+        while (node != null) {
+            n++;
+            int[] newItemset = insert(itemset, node.id);
+            if (list != null) {
+                synchronized (list) {
+                    list.add(new ItemSet(newItemset, support));
+                }
+            }
+            if (out != null) {
+                synchronized (out) {
+                    for (int i = 0; i < newItemset.length; i++) {
+                        out.format("%d ", newItemset[i]);
+                    }
+                    out.format("(%d)%n", support);
+                }
+            }
+            if (ttree != null) {
+                synchronized (ttree) {
+                    ttree.add(newItemset, support);
+                }
+            }
+
+            n += grow(out, list, ttree, node, newItemset, support);
+
+            node = node.parent;
+        }
+
+        return n;
+    }
+
+    /**
      * Mines FP-tree with respect to a single element in the header table.
      * @param header the header table item of interest.
      * @param itemset the item set represented by the current FP-tree.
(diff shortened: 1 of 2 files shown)


haifengl:
Thanks. Since you take out the code to a new function, the diff doesn't show the exact changes. Can you please explain what are exactly your changes?

peter-toth:
There is an optimization in the code to handle the case when the conditional FP-tree is a simple sequence of nodes. In that case we need to add all combinations of the items in the tree as frequent itemset. The original code was wrong with the nested loops, I changed it to a correct recursive solution.

haifengl:
Thanks. Have you try your changes on large datasets, especially with big itemsets? I chose non-recursive implement ion with the consideration of performance and also avoid of stack overflow. If we can keep non-recursive implementation, it would be great.

peter-toth:
FPGrowth mining is a recursive algo by its nature. But this optimization part can be implemented without recursion using loops and some stack data structure. I'm still benchmarking the non-recursive form, but unfortunatelly it is not obvious which one performs better.

haifengl:
What do you mean "not obvious"? On what data?

peter-toth:
I benchmarked the single path mining on datasets where there there is 24 (or more) nodes in the "single path" in the FP-tree. In this case 16.8M frequent itemsets are added to the result. The recursive version you can find in this PR and my non-recursive version (wip version can be found here: https://github.com/peter-toth/smile/commit/079c37dee8b28dfb4d55bf2f6c05b92c7cd763e8) performed very similarly. I also expected the non-recursive version to be faster, but it is not.
I'm will run some more benchmarks and share the results with you.
The advantage of the recursive version is the cleanness of the code.

peter-toth:
On this branch you can find the JMH benchmarks: https://github.com/peter-toth/smile/tree/benchmark-fp-growth-fix

In a propper benchmark the non-recursive version proved somewhat faster so I changed the implementation to that in this PR.

```
[info] Benchmark                                          Mode  Cnt   Score   Error  Units
[info] FPGrowthBenchmark.measureFPGrowth_20_NonRecursive  avgt   15   0,020 ▒ 0,001   s/op
[info] FPGrowthBenchmark.measureFPGrowth_20_Recursive     avgt   15   0,023 ▒ 0,001   s/op
[info] FPGrowthBenchmark.measureFPGrowth_24_NonRecursive  avgt   15   0,330 ▒ 0,011   s/op
[info] FPGrowthBenchmark.measureFPGrowth_24_Recursive     avgt   15   0,375 ▒ 0,014   s/op
[info] FPGrowthBenchmark.measureFPGrowth_28_NonRecursive  avgt   15   5,233 ▒ 0,102   s/op
[info] FPGrowthBenchmark.measureFPGrowth_28_Recursive     avgt   15   6,071 ▒ 0,118   s/op
[info] FPGrowthBenchmark.measureFPGrowth_30_NonRecursive  avgt   15  21,539 ▒ 0,303   s/op
[info] FPGrowthBenchmark.measureFPGrowth_30_Recursive     avgt   15  24,559 ▒ 0,512   s/op
```
Please see full logs here: https://github.com/peter-toth/smile/blob/benchmark-fp-growth-fix/benchmark/jmh.txt

haifengl:
Thank you!

PR #306, opened 2018-04-23
Support missing values in decision and regression tree

monapasan:
Double.NaN is a representation of a missing value.
Double.NaN is replaced with Double.MIN_VALUE to act as an initial value.

Splitting:
* if splitValue is not NaN, then put all of the NaN value into the branch that has a higher amount of examples in it.
* if splitValue is NaN put all of NaNs into a true child and all of the others into false child
* this way of handling missing values is according to the recommendations of Breiman et.al (1984) and can be achieved in R(`rpart`) with passing `usesurrogate=2`.

Ordering numerical feature:
* NaN values will be last in the ordering.

### DIFF
--- a/core/src/main/java/smile/classification/DecisionTree.java
+++ b/core/src/main/java/smile/classification/DecisionTree.java
@@ -277,7 +277,7 @@ public class DecisionTree implements SoftClassifier<double[]>, Serializable {
          */
         int splitFeature = -1;
         /**
-         * The split value.
+         * The split value. MIN_VALUE is just an initial value without further meaning.
          */
         double splitValue = Double.MIN_VALUE;
         /**
@@ -599,6 +599,7 @@ public class DecisionTree implements SoftClassifier<double[]>, Serializable {
                 }
             } else if (attributes[j].getType() == Attribute.Type.NUMERIC) {
                 int[] trueCount = new int[k];
+				// MIN_VALUE is an initial value without further meaning
                 double prevx = Double.MIN_VALUE;
                 int prevy = -1;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -114,7 +113,7 @@
     /**
      * The root of the regression tree
      */
-    private Node root;
+    public Node root;
```
haifengl: no, no, no.

core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -577,7 +586,7 @@ public Node findBestSplit(int n, int[] count, int[] falseCount, double impurity,
 
                     int trueLabel = Math.whichMax(trueCount[l]);
                     int falseLabel = Math.whichMax(falseCount);
-                    double gain = impurity - (double) tc / n * impurity(trueCount[l], tc) - (double) fc / n * impurity(falseCount, fc);
+					double gain = impurity - (double) tc / n * impurity(trueCount[l], tc) - (double) fc / n * impurity(falseCount, fc);
```
haifengl: With missing values, can we really still use this formula?

haifengl:
Thanks. But as I explained before, it may not be a good idea to incorporate such a missing value handling feature. It is not standard and many mathematical results (e.g. split metrics) may not hold any more given your changes.

core/src/main/java/smile/classification/DecisionTree.java
monapasan: This was needed for the application we use the library from as we need all of the information relating the tree. I can change it.

core/src/main/java/smile/classification/DecisionTree.java
monapasan: The question I also had in mind. I haven't found the formula in literature... Maybe you can give me a hint on where I can read about it?

monapasan:
@haifengl Thank you for taking a look at my changes!
In case there is not missing values in features, it should give exactly the same results. Regarding split metrics, I would love to hear your opinion why they can be corrupted.

haifengl:
There are many research in this area but I am not very familiar.

core/src/main/java/smile/classification/DecisionTree.java
monapasan: @haifengl any idea where I can find out about this formula?

haifengl: Any book or papers on decision tree or CART should have it.

PR #314, opened 2018-05-21
Improve KModes Model + check KModes refactor + add unit test

beckgael:
I improved speed of the model by changing sortBy(xxx).head by minBy(xxx).
I checked the Kmodes algorithms and i would say that what you've done is more like imperative style than functional contrary at what i've done which is in my sense is purely functional, maybe i'm wrong, please let me know.
From what i know, (foreach, map, flatMap) style is functional compared to for loop which is imperative style.

I will soon add the unit test.


haifengl:
For functional programming, I try to minimize the usage of `var`. `for` loop is just a syntactic sugar in Scala, which is translated to `map` or `foreach` by compiler. Thanks.

haifengl:
`KModesModel` is actually not used any more. I will remove it.

Issue #313, opened 2018-05-19
Fix spelling of DBSCAN


kno10:
The DBSCAN algorithm is incorrectly spelled "DBScan" throughout smile.

haifengl:
Since you care about it, can you fix and make a pull request? Thanks.

kno10:
I don't actually use Smile.

As this will break APIs when renaming the class, it is probably something you should decide how you want to handle a transition. E.g. by renaming the class to DBSCAN but adding a DBScan legacy subclass instead. Or just postpone this to the next major release.

Just the documentation change is of course an easy s/DBScan/DBSCAN/ - except for class names...

PR #318, opened 2018-05-29
Rename DBScan to DBSCAN

kno10:
Full fix for #313, but no compatibility wrapper for the old class name.

Includes PR #317


haifengl:
Thanks

Issue #320, opened 2018-06-11
LDA, FLD, QDA algorithm implementation references


rayeaster:
@haifengl could please shed some light on following questions? Really appreciate your time.

1) in smile.classification.LDA, following code snippet calculate the cross variance matrix, is there any reason or reference about the usage of **_n-k_** at line 287?

![image](https://user-images.githubusercontent.com/1487283/41228964-d9b008fa-6dac-11e8-9a8c-ae52ac839b72.png)

and there exist some magic number for prediction, again, any reference would be great to understand.
![image](https://user-images.githubusercontent.com/1487283/41231296-37e86752-6db5-11e8-8199-958e301f343c.png)

2) in smile.classification.FLD, following code snippet calculate the desirable projection matrix, is there any reason or reference about the usage of code between line 289 and line 297? 

And It seems that the testing result on demo even better with those codes removed. Current version of code would result in a projection matrix directly from between class scatter matrix...

From smile documentation, the implementation seems not correct. But I tried `U.inverse().abmm(B)` which give worse result on demo data set than current implementation
![image](https://user-images.githubusercontent.com/1487283/41345842-54b63bba-6f37-11e8-889f-2eaac52fd021.png)

![image](https://user-images.githubusercontent.com/1487283/41229426-8c21801c-6dae-11e8-9f30-17219197920d.png)

3) in smile.classification.QDA, following code snippet calculate the discriminant function for classes(which is quite different from LDA), is there any reason or reference about the calculation?
![image](https://user-images.githubusercontent.com/1487283/41232115-daff1fce-6db7-11e8-8605-6ac143d67082.png)

haifengl:
Check this paper https://ieeexplore.ieee.org/abstract/document/1498032/

rayeaster:
@haifengl 
that is great reference and Thanks. Thought we could add the reference in the code for clarity. 
BTW, I checked the paper and submit a PR [here ](https://github.com/haifengl/smile/pull/322) you want to have a look at. 
Really appreciate your time.

rayeaster:
another minor issue is about the prediction formula `f = ct[i] - 0.5 * f;` in LDA and QDA which I failed to find reference in the paper. Could you please mind to point it out? 
@haifengl

haifengl:
LDA takes class labels when calculating `C`. PCA doesn't.

haifengl:
For more details, check the book `Pattern Classification` by Duda, Hart, & Stork.

PR #322, opened 2018-06-14
update fld training with regard to given reference

rayeaster:
This change reflect the reference paper implementation.

http://alumni.cs.ucr.edu/~hli/paper/hli05tumor.pdf

Test result on demo data set shows the change has a better error rate with a smooth decision boundary. Check related issues [here](https://github.com/haifengl/smile/issues/320)

image

old demo error rate
`training error = 21.00% for toy and training error = 20.14% for big toy`

new demo error rate
`training error = 19.00% for toy and training error = 18.73% for big toy`

image

image


haifengl:
Thanks. What about other datasets, e.g. unit tests? The demo is only for, well, "demo". They are not serious test cases.

core/src/main/java/smile/classification/FLD.java
```diff
@@ -273,32 +278,51 @@ public FLD(double[][] x, int[] y, int L, double tol) {
...
-        }
-
+        }     
+        //
+        // calculate the square root of diagonal inverse of eigenvalue matrix from SVD of covariance matrix
+        //
         DenseMatrix U = eigen.getEigenVectors();
-        DenseMatrix UB = U.atbmm(B);
-
+        DenseMatrix D = eigen.getD().inverse();
```
haifengl: This is slow and also not numerically stable.  D is a diagonal matrix, which may have zeros in the diagonal. But `inverse()` is a general method, suitable for dense matrix.

core/src/main/java/smile/classification/FLD.java
```diff
@@ -273,32 +278,51 @@ public FLD(double[][] x, int[] y, int L, double tol) {
...
+        EVD eigen = T.eigen();  
         
         tol = tol * tol;
         double[] s = eigen.getEigenValues();
         for (int i = 0; i < s.length; i++) {
             if (s[i] < tol) {
                 throw new IllegalArgumentException("The covariance matrix is close to singular.");
             }
-
-            s[i] = 1.0 / s[i];
```
haifengl: why remove this?

rayeaster:
Good point . I could try more dataset or any perticular dataset you have on mind

core/src/main/java/smile/classification/FLD.java
rayeaster: This is from the reference paper if what I understand is correct.
image

core/src/main/java/smile/classification/FLD.java
rayeaster: not quite clear about the usage of this line

rayeaster:
And I think current LDA implementation is actually PCA style

core/src/test/java/smile/classification/NaiveBayesTest.java
```diff
@@ -266,7 +344,7 @@ public void testLearnBernoulli() {
             }
         }
 
-        System.out.format("Bernoulli error = %d of %d%n", error, total);
```
haifengl: `%n` is a portal way for `\n`. Please don't change this.

core/src/test/java/smile/classification/NaiveBayesTest.java
```diff
@@ -307,7 +385,7 @@ public void testLearnBernoulli2() {
...
```
haifengl: same above.

core/src/test/java/smile/classification/NaiveBayesTest.java
```diff
@@ -101,7 +101,7 @@ public void tearDown() {
      */
```
haifengl: Can we separate this file change from the above? They are not related topics.

haifengl:
There is no `U.abmm(U.atbmm(B))`. You missed the lines 291-295. We actually calculate inv(S_t) * S_b.

core/src/test/java/smile/classification/NaiveBayesTest.java
rayeaster: Yes I will keep it as it was. Thanks for pointing out

haifengl:
It is very expensive and instable to calculate inv(T) directly. In general, it is not a good idea to translate your math formula to code directly. Computational mathematics studies how to calculate them quickly and reliably. It is the gap you missed.

haifengl:
For all the data you tested, we should get the same results with the old implementation and new one because the data do not in the small sample size problem as discussed in the paper. The difference may be caused by the numerical stability. Iris has only 4 variables. So it is probably better that we use the old implementation that works on T (4x4 matrix, very small). For other data with high dimensions, we should use the fast algorithm/new implementation.

Of course, the difference may also caused by some bug :) I will take some time to review both implementations. Thanks!

rayeaster:
So which do you prefer? Keep current implementation as it is or you want to see more tests before considering new implementation:)

rayeaster:
That sounds great:) Thank you for taking time to review this. Maybe we could use a parmeter to switch between 2 implementations for different data size. Also a lot thanks for the comment about computational math.BTW any resource on that topic you recommend as reference?

haifengl:
Yes, it is what I would like.

rayeaster:
Great. I could add the parameter

rayeaster:
@haifengl do you have time to review the implementation difference. Thanks.

rayeaster:
@haifengl please let me know if you see any implementation mistake in this Pull Request or you spot any inappropriate coding style. Thanks.

> 10-fold

oss

haifengl:
I will cherry pick your PR. I still believe that two algorithms should generate the same model for the test data in the unit tests. These data don't have the small sample size problem. So the two algorithms should be same mathematically.

rayeaster:
@haifengl maybe because the old implementation calculate `inv(S_t) * S_b`, not `inv(S_w) * S_b`?

haifengl:
Mathematically, `inv(S_t) * S_b` and `inv(S_w) * S_b` are same in sense of optimizing the FLD criterion when `S_w` is not singular.

rayeaster:
so what should be the cause for discrepancy there?

haifengl:
I have cherry-picked this into v2. Thanks!

PR #325, opened 2018-06-27
Fixtypo

rayeaster:
This PR fix some minor typo in java doc for OLS and SVD

### DIFF
--- a/math/src/main/java/smile/math/matrix/SVD.java
+++ b/math/src/main/java/smile/math/matrix/SVD.java
@@ -253,7 +253,7 @@ public class SVD {
     /**
      * Solve the least squares A*x = b.
      * @param b   right hand side of linear system.
-     * @param x   the output solution vector that minimizes the L2 norm of U*x - b.
+     * @param x   the output solution vector that minimizes the L2 norm of A*x - b.
      * @exception  RuntimeException if matrix is rank deficient.
      */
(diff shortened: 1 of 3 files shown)


rayeaster:
add a refer for QR/SVD: https://www.quora.com/Is-it-better-to-do-QR-Cholesky-or-SVD-for-solving-least-squares-estimate-and-why

math/src/main/java/smile/math/matrix/SVD.java
```diff
@@ -253,7 +253,7 @@ public Cholesky CholeskyOfAtA() {
     /**
      * Solve the least squares A*x = b.
      * @param b   right hand side of linear system.
-     * @param x   the output solution vector that minimizes the L2 norm of Q*R*x - b.
+     * @param x   the output solution vector that minimizes the L2 norm of U*x - b.
```
haifengl: why U? Isn't it A?

rayeaster: For DenseMatrix.svd():  

Returns the singular value decomposition. Note that the input matrix will hold U on output.

and in QR.solve() javadoc, it use QR as specific explanation instead of A

rayeaster: Got it. updated.

haifengl:
Thanks

PR #332, opened 2018-07-13
add sgd for Logistic Regression

rayeaster:
add stochastic gradient descent for Logistic Regression
reference: http://www.cs.cmu.edu/~wcohen/10-605/sgd-part2.pdf

### DIFF
--- a/core/src/main/java/smile/classification/LogisticRegression.java
+++ b/core/src/main/java/smile/classification/LogisticRegression.java
@@ -75,7 +75,7 @@ import smile.util.MulticoreExecutor;
  * 
  * @author Haifeng Li
  */
-public class LogisticRegression implements SoftClassifier<double[]>, Serializable, OnlineClassifier<double[]> {
+public class LogisticRegression implements SoftClassifier<double[]>, OnlineClassifier<double[]>, Serializable {
     private static final long serialVersionUID = 1L;
     private static final Logger logger = LoggerFactory.getLogger(LogisticRegression.class);
 
@@ -112,7 +112,7 @@ public class LogisticRegression implements SoftClassifier<double[]>, Serializabl
     /**
      * learning rate for stochastic gradient descent.
      */
-    double step = 5e-5;
+    double eta = 5e-5;
 
     /**
      * Trainer for logistic regression.
@@ -851,9 +851,9 @@ public class LogisticRegression implements SoftClassifier<double[]>, Serializabl
 
 			double[] g = new double[p + 1];
 			for (int j = 0; j < p; j++) {
-				g[j] = step * res * x[j];
+				g[j] = eta * res * x[j];
 			}
-			g[p] = step * res;
+			g[p] = eta * res;
 
 			// update the weights
 			for (int j = 0; j <= p; j++) {
@@ -861,7 +861,7 @@ public class LogisticRegression implements SoftClassifier<double[]>, Serializabl
 
 				// add regularization part
 				if (lambda != 0.0) {
-					w[j] -= 2 * lambda * step * w[j];
+					w[j] -= 2 * lambda * eta * w[j];
 				}
 			}
 		} else {
@@ -878,9 +878,9 @@ public class LogisticRegression implements SoftClassifier<double[]>, Serializabl
 				yi = (y == j ? 1.0 : 0.0) - prob[j];
 
 				for (int l = 0; l < p; l++) {
-					g[j][l] = step * yi * x[l];
+					g[j][l] = eta * yi * x[l];
 				}
-				g[j][p] = step * yi;
+				g[j][p] = eta * yi;
 			}
 
 			// update the weights
@@ -890,7 +890,7 @@ public class LogisticRegression implements SoftClassifier<double[]>, Serializabl
 
 					// add regularization part
 					if (lambda != 0.0) {
-						W[j][l] -= 2 * lambda * step * W[j][l];
+						W[j][l] -= 2 * lambda * eta * W[j][l];
 					}
 				}
 			}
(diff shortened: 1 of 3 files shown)


rayeaster:
unit test result:

```
Segment
Segment error rate = 5.93%
SGD-Segment
SGD-Segemnt one more epoch...
SGD-Segment error rate = 6.17%

Iris binary
Logistic Regression error = 3
Iris
Logistic Regression error = 3
Iris binary
SGD-Iris Binary one more epoch...
SGD-Iris Binary Logistic Regression error = 3

USPS
USPS error rate = 9.37%
SGD-USPS
SGD-USPS one more epoch...
SGD-USPS error rate = 9.37%

```

haifengl:
Thanks for quickly implementing this. It is interesting that the error rate is higher after one more epoch. Shall we shuffle the samples first? May be the samples are ordered by their class labels. It is not good for SGD.

core/src/main/java/smile/classification/LogisticRegression.java
```diff
@@ -823,7 +834,78 @@ public double f(double[] w, double[] g) {
...
+				}
+			}
+		}
+	}	
+
+	/**
+	 * Tune the fixed learning rate for stochastic gradient descent
+	 * @param step usually quite small to avoid oscillation, default is 5e-5
+	 */
+    public void setStep(double step) {
```
haifengl: `setLearningRate()` is better. THe learning rate in NeuralNetwork is called `eta`. Let's be consistent.

math/src/main/java/smile/math/Math.java
```diff
@@ -4411,7 +4411,7 @@ public static double min(DifferentiableMultivariateFunction func, int m, double[
         // Current function value.
         double f = func.f(x, g);
 
-        logger.info(String.format("L-BFGS: initial function value: %.5g", f));
+        logger.debug(String.format("L-BFGS: initial function value: %.5g", f));
```
haifengl: Why `debug`? I prefer see the messages to asses if the algorithm is converging.

core/src/main/java/smile/classification/LogisticRegression.java
```diff
@@ -75,7 +75,7 @@
  * 
  * @author Haifeng Li
  */
-public class LogisticRegression implements SoftClassifier<double[]>, Serializable {
+public class LogisticRegression implements SoftClassifier<double[]>, Serializable, OnlineClassifier<double[]> {
```
haifengl: Please put `OnlineClassifier` before `Serializable`.

core/src/main/java/smile/classification/LogisticRegression.java
rayeaster: so setLearningRate() or setEta()?

core/src/main/java/smile/classification/LogisticRegression.java
rayeaster: consider it done.

math/src/main/java/smile/math/Math.java
rayeaster: ok. I will revert this change.

rayeaster:
"Shall we shuffle the samples first?"
any utility class I could use for sample shuffle? @haifengl

core/src/main/java/smile/classification/LogisticRegression.java
haifengl: use `setLearningRate` for the method name and use `eta` for the variable name. Thanks.

haifengl:
There is `permutate` method in `smile.math.Math`

core/src/main/java/smile/classification/LogisticRegression.java
rayeaster: Thanks and got it.

rayeaster:
@haifengl  code changed according your input and unit test result remains the same. Thanks,

core/src/main/java/smile/classification/LogisticRegression.java
```diff
@@ -823,7 +834,78 @@ public double f(double[] w, double[] g) {
...
+		if (x.length != p) {
+			throw new IllegalArgumentException("Invalid input vector size: " + x.length);
+		}
+
+		if (k == 2) {
+			// calculate gradient for incoming data
+			double wx = dot(x, w);
+			double res = y - Math.logistic(wx);
+
+			double[] g = new double[p + 1];
```
haifengl: no need to create a new an array here. It introduces memory fragment and slows down the system when we update the model with a lot of samples. It is easy to combine this loop with the below one and just use a local variable.

core/src/main/java/smile/classification/LogisticRegression.java
```diff
@@ -823,7 +834,78 @@ public double f(double[] w, double[] g) {
...
+			}
+		} else {
+			double[] prob = new double[k];
+			for (int j = 0; j < k; j++) {
+				prob[j] = dot(x, W[j]);
+			}
+
+			softmax(prob);
+
+			double[][] g = new double[k][p + 1];
```
haifengl: same as above comment.

haifengl:
I think that the unit test with SGD should be revised. Note that our batch learning reaches the global optimal. By online learning with the same data, we won't get better results (actually could be worse). I suggest that we train with only part of data (says 80% samples) with batch algorithm. And then update the model with online learning with the rest of samples.

rayeaster:
unit test results with 80% samples for batch algorithm, 20% training samples for sgd (shuffle enabled) : 
SGD-Segment error rate = 6.17%
SGD-Iris Binary Logistic Regression error = 8
SGD-USPS error rate = 9.57%

unit test results with optimal error rate for various parameter-tuning I tried : 
SGD-Segment error rate = **5.68**% （**70**% samples for batch algorithm）
SGD-Iris Binary Logistic Regression error = **3**（**95**% samples for batch algorithm）
SGD-USPS error rate = **9.57**%（**80**% samples for batch algorithm）

haifengl:
Thanks!

rayeaster:
my pleasure

PR #335, opened 2018-07-16
Upgrade SBT and use https for resolvers

benmccann:
It's a security issue to use http for a resolver. The jars you download could be switched out for malicious versions. See here: https://max.computer/blog/how-to-take-over-the-computer-of-any-java-or-clojure-or-scala-developer/


haifengl:
Thanks

Issue #297, opened 2018-04-01
Feature: variable naming


benmccann:
It would be nice to optionally allow variables to be named. E.g. when calling `OLS.toString` the output could be made much more user friendly

Spark DataFrames allow this for example

haifengl:
Will do.

haifengl:
See PR #336

PR #336, opened 2018-07-16
Include variable names for OLS

benmccann:
Partially addresses https://github.com/haifengl/smile/issues/297. There are probably other algorithms where this would be useful as well

I implemented this with a simple `String[]`. However, it may be preferable to use a dataframe like [Tablesaw](https://github.com/jtablesaw/tablesaw) similar to Spark's dataframe. If you would like, I could send a PR to utilize a dataframe as a nicer datastructure for this information.

### DIFF
--- a/core/src/main/java/smile/regression/OLS.java
+++ b/core/src/main/java/smile/regression/OLS.java
@@ -18,8 +18,12 @@ package smile.regression;
 
 import java.io.Serializable;
 import java.util.Arrays;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import smile.data.Attribute;
+import smile.data.AttributeDataset;
 import smile.math.Math;
 import smile.math.matrix.Matrix;
 import smile.math.matrix.DenseMatrix;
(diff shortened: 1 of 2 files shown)


haifengl:
We have `AttributeDataset` that provide attribute name and other info. RandomForest, for example, takes it as input.

benmccann:
I don't think `RandomForest` does take `AttributeDataset` as input actually. I can only find tests that use `AttributeDataset`

benmccann:
Oh, it does take `Attribute[] attributes` as input though

benmccann:
I will update this PR this evening

benmccann:
@haifengl thanks for the helpful pointer! I've updated this PR so that `OLS` takes an `AttributeDataset`

haifengl:
Thanks!

haifengl:
Unit test fails with this change. I fixed it. Please make sure to run unit tests first for PRs. Thanks.

benmccann:
Thank you! Sorry for the breakage

PR #337, opened 2018-07-17
Add constructors taking AttributeDataset

benmccann:
AttributeDataset is often much easier to construct because you can convert arff files, Tablesaw tables, etc. to `AttributeDataset`

We probably don't need constructors taking `Attribute[] attributes, double[][] x, int/double[] y` as it's easier to just pass an `AttributeDataset` but I left them for backwards compatibility


haifengl:
Thanks

Issue #217, opened 2017-08-30
Allow for arbitrary column to contain row names/IDs


DewaldDeJager:
This is specific to the `DelimitedTextParser` and it's `hasRowNames` attribute. Currently the class accepts a boolean to indicate whether the first column contains row IDs. It may be useful to accept an integer instead and have this indicate the column index of the column containing the row IDs (Similar to response index). This would purely be for convenience and may avoid some preprocessing.

haifengl:
In v2, we support DataFrame, which may include a column as row name. When fitting a model, use Formula to select variables for fitting.

PR #341, opened 2018-08-02
Allow for arbitrary column to contain row names/IDs

anuj1207:
Using integer instead of boolean value for row names column index
and this integer points to the column index where we'll get the row names
Link to the issue: https://github.com/haifengl/smile/issues/217

### DIFF
--- a/data/src/main/java/smile/data/parser/DelimitedTextParser.java
+++ b/data/src/main/java/smile/data/parser/DelimitedTextParser.java
@@ -62,9 +62,9 @@ public class DelimitedTextParser {
      */
     private boolean hasColumnNames = false;
     /**
-     * The dataset has row names at first column.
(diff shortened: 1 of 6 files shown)


haifengl:
Thanks. Although it is more flexible, we prefer a simpler API and compatibility with R. It is rare that people has header in other rows.

anuj1207:
@haifengl So this issue is not needed?? Are we moving ahead on this one or should I close it?

haifengl:
It is hard to find a real need for this. If the first row is not the header, it is unlikely the data either. Probably other metadata. It is hard to develop a general method for text/csv files with multi-row metadata. We have other dedicated methods for these cases, e.g. arff and a variety of gene expression data. I would like to keep the text/csv parser simple. Thanks.

Issue #347, opened 2018-08-21
fastICA implementation


rayeaster:
typical use case is 'cocktail party problem' where to identify original speakers from mixed audio recordings:

https://www.cs.helsinki.fi/u/ahyvarin/papers/TNN99new.pdf
https://en.wikipedia.org/wiki/FastICA

haifengl:
They are not really machine learning algorithm. But if you want, you may implement it in the `math` module.

rayeaster:
glad to help:) But why not add it to projection package in core like PCA?

haifengl:
Maybe. Can it support the methods of `Projection` interface?

rayeaster:
of course:)

haifengl:
ICA is not a projection. It is wrong to use the final result as a projection matrix. Besides, there are several mathematical mistakes in your implementation (maybe confused by the signal/source, i.e. row/column). Thanks

haifengl:
BTW, you should avoid matrix multiplication route (N^3) when one matrix is diagonal. It can be done in O(N^2).

PR #349, opened 2018-08-22
fastica

rayeaster:
add fastica in projection, refers to https://github.com/haifengl/smile/issues/347


rayeaster:
image

image

haifengl:
Can you add documents/comments to all methods? Thanks.

haifengl:
ICA is good for cocktail party problem. Do the data in your demo/test make sense for it?

haifengl:
Can we rename the class name to `ICA`? The interface should reflect the purpose only. FastICA is only one way to do the computation. The user should not bother the implementation details. Thanks.

rayeaster:
> Can you add documents/comments to all methods? Thanks.

will do

rayeaster:
> Can we rename the class name to `ICA`? The interface should reflect the purpose only. FastICA is only one way to do the computation. The user should not bother the implementation details. Thanks.

got it. will do

rayeaster:
> ICA is good for cocktail party problem. Do the data in your demo/test make sense for it?

The demo follow same setting with other projection mechanism like PCA.
The unit test indeed do not use specific data set from cocktail problem or mixed signals. If time permit, I would like to add a similar data set of that type. Any dataset you have at hands for this purpose?

haifengl:
I don't have test for cocktail problem. There should be some standard benchmark data for ICA. It is also a good opportunity for you to compare the correctness of the implementation on the benchmark data.

rayeaster:
@haifengl  modified according to your comments above. 

Add a mixed signal test dataset from sklearn: 
http://scikit-learn.org/stable/auto_examples/decomposition/plot_ica_blind_source_separation.html

Result from this PR (ICA) and smile PCA comes at the 3th and 4th pictures from top:
image

also include the reference pseudo-code here:

image
image

haifengl:
Thanks. Do you use tab in the code? I observe it in your commits across multiple PR. Can you set tab -> space in your IDE and also set indent to 4?

rayeaster:
> Thanks. Do you use tab in the code? I observe it in your commits across multiple PR. Can you set tab -> space in your IDE and also set indent to 4?

sure. will do. sorry had not noticed that.

core/src/test/java/smile/projection/PCATest.java
```diff
@@ -272,4 +279,56 @@ public void testPCACor() {
             }
         }
     }
+
```
haifengl: This is not necessary. PCA and ICA are for different purposes. And I don't see you really test any real results here.

core/src/test/java/smile/projection/ICATest.java
```diff
@@ -0,0 +1,183 @@
...
+
+    @After
+    public void tearDown() {
+    }
+
+    /**
+     * Test of learn method, of FastICA.
+     */
+    @Test
+    public void testFastICA() {
```
haifengl: no need to test on iris. the data doesn't make sense for ICA.

rayeaster:
@haifengl thanks for this thorough code review. I modify the PR with respect to your comments and add test correctness check. It now shows comparable (even better) result with sklearn package (updated plot is embed in comments above):

smile:
rmse for signal1 is 0.556332788860372
rmse for signal2 is 0.502187993412292
rmse for signal3 is 0.5816896466939584

sklearn:
rmse for signal1 is 0.70226691359
rmse for signal2 is 0.999406929533
rmse for signal3 is 0.579000347791

core/src/main/java/smile/projection/ICA.java
```diff
@@ -0,0 +1,365 @@
...
+                T.set(l, j, T.get(j, l));
+            }
+        }
+
+        T.setSymmetric(true);
+        EVD eigen = T.eigen();
+        DenseMatrix D = eigen.getEigenVectors();
+        double[] v = eigen.getEigenValues();
+        for (int i = 0; i < v.length; i++) {
+            v[i] = 1 / Math.sqrt(v[i]);
```
haifengl: we need to safeguard `v[i]` is not too small or zero.

core/src/main/java/smile/projection/ICA.java
```diff
@@ -0,0 +1,365 @@
...
+        }
+        DenseMatrix V = DenseMatrix.diag(v);
+        DenseMatrix white = D.abmm(V).abtmm(D);
+
+        DenseMatrix ct = white.abtmm(centered);
```
haifengl: why `abtmm` and then `transpose` in the blow? Isn't it just `white.abmm(X)`?

core/src/main/java/smile/projection/ICA.java
```diff
@@ -0,0 +1,365 @@
...
+        for (int j = 0; j < n; j++) {
+            for (int l = 0; l <= j; l++) {
+                T.div(j, l, m);
+                T.set(l, j, T.get(j, l));
+            }
+        }
+
+        T.setSymmetric(true);
+        EVD eigen = T.eigen();
+        DenseMatrix D = eigen.getEigenVectors();
```
haifengl: Why `D` not `E` as in the wiki link you post? Keeping variable names consistent with your reference will make it easy to check the formula.

core/src/main/java/smile/projection/ICA.java
```diff
@@ -0,0 +1,365 @@
...
+
+    /**
+     * determine the functions to calculate derivative and second-order derivative
+     */
+    private NegEntropyFunc funcMode = NegEntropyFunc.LOGCOSH;
+
+    /**
+     * functions to calculate derivative and second-order derivative
+     */
+    public enum NegEntropyFunc {
```
haifengl: It is better to define an interface that has `f` (nonquadratic nonlinear function), `g` (first derivative) and `g2` (second derivative). And provide implementations for `log cosh` and `- exp(-u^2/2)`. This way the user can easily plug in their special functions without changing this code.

haifengl: Make this interface and `log cosh` etc implementations inner classes of ICA.

core/src/main/java/smile/projection/ICA.java
```diff
@@ -0,0 +1,365 @@
...
+     *            target independent component count
+     * @param tol
+     *            convergence threshold on component search
+     * @param maxIter
+     *            maximum iteration on component search
+     * @param negfunc
+     *            see {@link NegEntropyFunc}
+     */
+    public ICA(double[][] data, int p, double tol, int maxIter, NegEntropyFunc negfunc) {
+        int m = data.length;
```
haifengl: hard to follow `m` and `n` as they are swapped compared the wiki page.

core/src/main/java/smile/projection/ICA.java
```diff
@@ -0,0 +1,365 @@
...
+
+        projection = Matrix.zeros(p, n);
+
+        GaussianDistribution g = new GaussianDistribution(0, 1);
+        double[][] wps = new double[p][n];
+        for (int i = 0; i < p; i++) {
+            for (int j = 0; j < n; j++) {
+                wps[i][j] = g.rand();
+            }
+            // normalize
```
haifengl: why normalization? `w` is already standard normal distributed.

core/src/main/java/smile/projection/ICA.java
```diff
@@ -0,0 +1,365 @@
...
+     *            convergence threshold on component search
+     * @param maxIter
+     *            maximum iteration on component search
+     * @param negfunc
+     *            see {@link NegEntropyFunc}
+     */
+    public ICA(double[][] data, int p, double tol, int maxIter, NegEntropyFunc negfunc) {
+        int m = data.length;
+        this.n = data[0].length;
+        this.p = p;
```
haifengl: check `p <= n`.

core/src/main/java/smile/projection/ICA.java
```diff
@@ -0,0 +1,365 @@
...
+                wps[i][j] = g.rand();
+            }
+            // normalize
+            double norm = Math.norm2(wps[i]);
+            for (int j = 0; j < n; j++) {
+                wps[i][j] /= norm;
+            }
+        }
+
+        DenseMatrix whitened = whiteData(data);
```
haifengl: rename `whitened` to `X`. The expression will be shorter and easy to check correctness.

core/src/main/java/smile/projection/ICA.java
```diff
@@ -0,0 +1,365 @@
...
+        for (int i = 0; i < p; i++) {
+            double[] lastwp = wps[i];
+
+            double diff = Double.MAX_VALUE;
+            int iterCnt = 0;
+            while (diff > this.tolerance) {
+                // step 1 calculate derivative of projection from non-linear functions
```
haifengl: This part is too complicated compared to the simple formula in the reference. We can simply it greatly if leverage Matrix interface more.

core/src/main/java/smile/projection/ICA.java
```diff
@@ -0,0 +1,365 @@
...
+                double[] projSD = new double[m];
+                double sumProjSD = 0;
+                for (int idx = 0; idx < m; idx++) {
+                    projSD[idx] = funcMode == NegEntropyFunc.LOGCOSH ? getDeriv2WithLOGCOSH(proj[idx])
+                            : getDeriv2WithEXP(proj[idx]);
+                    sumProjSD += projSD[idx];
+                }
+
+                double[] wp = new double[n];
+                wp = whitened.transpose().ax(projFD, wp);
```
haifengl: `atx` is the right way

core/src/test/java/smile/projection/ICATest.java
```diff
@@ -0,0 +1,167 @@
...
+        parser.setResponseIndex(new NominalAttribute("class"), 3);
+        try {
+            AttributeDataset mixed = parser
+                    .parse(smile.data.parser.IOUtils.getTestDataFile("projection/mixed-signal/test_ica_X.csv"));
+            double[][] x = mixed.toArray(new double[mixed.size()][]);
+
+            parser.setResponseIndex(new NominalAttribute("class"), 1);
+            AttributeDataset signal1 = parser
+                    .parse(smile.data.parser.IOUtils.getTestDataFile("projection/mixed-signal/test_ica_s1.csv"));
+            double[][] s1 = signal1.toArray(new double[signal1.size()][]);// sinusoidal
```
haifengl: Are `s1`, `s2` and `s3` expected signals?

core/src/test/java/smile/projection/ICATest.java
```diff
@@ -0,0 +1,167 @@
...
+            double rmse2 = Math.sqrt(mse2 / size);
+            System.out.println("rmse for signal2 is " + rmse2);
+            assertTrue(rmse2 < 1);
+
+            double mse3 = 0;
+            for (int i = 0; i < size; i++) {
+                double diff = p3[i] - o1[i];
+                mse3 += diff * diff;
+            }
+            double rmse3 = Math.sqrt(mse3 / size);
```
haifengl: Not sure RMSE is the right metric to measure ICA.

rayeaster: @haifengl rmse here is used here as a quick and rough comparison taken point-by-point. Yes, for ICA, I think the most intuitive way is to check the visualization of unmixed-signals which also shows the power of ICA to distinguish mixed signals.

core/src/test/java/smile/projection/ICATest.java
rayeaster: @haifengl  yes, the original signal data. I extract the data from the code here: http://scikit-learn.org/stable/auto_examples/decomposition/plot_ica_blind_source_separation.html#sphx-glr-auto-examples-decomposition-plot-ica-blind-source-separation-py

core/src/main/java/smile/projection/ICA.java
rayeaster: will change.thanks

core/src/main/java/smile/projection/ICA.java
rayeaster: will change.thanks

core/src/main/java/smile/projection/ICA.java
rayeaster: will change.thanks

rayeaster:
@haifengl I will revise the PR with your current comments next week.

core/src/main/java/smile/projection/ICA.java
```diff
@@ -0,0 +1,365 @@
...
+        // center data
+        double[] mu = Math.colMeans(data);
+        DenseMatrix centered = Matrix.newInstance(data);
+        for (int j = 0; j < n; j++) {
+            for (int i = 0; i < m; i++) {
+                centered.sub(i, j, mu[j]);
+            }
+        }
+
+        // covariance matrix on centered data.
```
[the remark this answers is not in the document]
rayeaster: > In fact, you can do a `centered.aat()` here for the covariance matrix as it is already centered.

`centered.aat()`  that not working

core/src/main/java/smile/projection/ICA.java
rayeaster: > why normalization? `w` is already standard normal distributed.

each value in w is normal distributed, but w itself as a vector is not normalized. I borrow code from `Math.randn`

rayeaster: > why normalization? `w` is already standard normal distributed.

each value in w is normal distributed, but w itself as a vector is not normalized. I borrow code from `Math.randn`

bklemme:
I would really like to be able to use this new feature. Is there a plan to merge and/or release it soon? If it needs additional work I would be willing to take it on in order to facilitate the release.

haifengl:
It will be in next release.

bklemme:
Thanks! Any idea of the time frame for the next release?

haifengl:
I have added ICA to v2. Thanks

rayeaster:
> I have added ICA to v2. Thanks

I will close this PR then. Thanks

Issue #354, opened 2018-09-05
add Isotonic Regression scaling for classification posteriori propability


rayeaster:
> An alternative approach to probability calibration is to fit an isotonic regression model to an ill-calibrated probability model. This has been shown to work better than Platt scaling, in particular when enough training data is available --- from Platt Scaling wiki page

https://www.cs.cornell.edu/~alexn/papers/calibration.icml05.crc.rev3.pdf

rayeaster:
PR is here : https://github.com/haifengl/smile/pull/356

rayeaster:
@haifengl and would you mind to share the reference for static method of `PlattScaling#multiclass()` which looks quite amazing?

haifengl:
Probabilistic Outputs for Support Vector Machines and Comparisons to
Regularized Likelihood Methods

https://www.researchgate.net/profile/John_Platt/publication/2594015_Probabilistic_Outputs_for_Support_Vector_Machines_and_Comparisons_to_Regularized_Likelihood_Methods/links/004635154cff5262d6000000.pdf

rayeaster:
Really appreciate!

PR #356, opened 2018-09-06
add isotonic regression scaling to SVM

rayeaster:
add IsotonicRegressionScaling as a new probability calibration method
https://github.com/haifengl/smile/issues/354

http://fastml.com/classifier-calibration-with-platts-scaling-and-isotonic-regression/
https://en.wikipedia.org/wiki/Isotonic_regression#Algorithms

image

### DIFF
--- a/core/src/main/java/smile/classification/IsotonicRegressionScaling.java
+++ b/core/src/main/java/smile/classification/IsotonicRegressionScaling.java
@@ -37,6 +37,9 @@ import smile.sort.QuickSort;
  */
 public class IsotonicRegressionScaling implements Serializable {
     private static final long serialVersionUID = 1L;
+    
+    private double trainMixScore = Double.NaN;
+    private double trainMaxScore = Double.NaN;
 
     /** setep-wise constant functions */
     private static class StepwiseConsFunc {
@@ -71,7 +74,7 @@ public class IsotonicRegressionScaling implements Serializable {
         }
 
         boolean matchValue(double score) {
-            if (score > lowScore && score <= highScore) {
+            if ((score - lowScore) > 0.0 && (highScore - score) >= 0.0) {
                 return true;
             } else {
                 return false;
@@ -97,11 +100,15 @@ public class IsotonicRegressionScaling implements Serializable {
         initStepwiseConstFuncs(scores, y);
 
         StepwiseConsFunc current = startConstantFuncs.next;
-
+        StepwiseConsFunc last = startConstantFuncs;
+        
         while (current != null) {
             StepwiseConsFunc previous = current.previous;
             while (previous != null) {
-                if (previous.getVal() >= current.getVal()) {
+                double pv = previous.getVal();
+                double cv = current.getVal();
+                double diff = pv - cv;
+                if (diff >= 0.0) {
                     int weight = previous.getWeight() + current.getWeight();
                     double val = (previous.getWeight() * previous.getVal() + current.getWeight() * current.getVal())
                             / weight;
@@ -109,23 +116,35 @@ public class IsotonicRegressionScaling implements Serializable {
                     newFunc.setWeight(weight);
 
                     newFunc.next = current.next;
-                    if (newFunc.next != null) {
+                    if (current.next != null) {
                         newFunc.next.previous = newFunc;
+                        current.next = null;
                     }
 
                     newFunc.previous = previous.previous;
-                    if (newFunc.previous != null) {
+                    if (previous.previous != null) {
                         newFunc.previous.next = newFunc;
+                        previous.previous = null;
                     }
-
+                    
                     current = newFunc;
                     previous = current.previous;
+                    if (previous == null) {
+                        startConstantFuncs = current;
+                    }
                 } else {
                     previous = previous.previous;
                 }
             }
+            last = current;
             current = current.next;
         }
+        
+        double minScore = startConstantFuncs.lowScore;
+        assert((minScore - trainMixScore) == 0.0);
+        
+        double maxScore = last.highScore;
+        assert((maxScore - trainMaxScore) == 0.0);
     }
 
     private void initStepwiseConstFuncs(double[] scores, int[] y) {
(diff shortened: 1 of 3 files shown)


rayeaster:
Note that usually it is recommended to use separate data set other than traing data set for probability calibration, i.e., probability calibration should be done on new data not used for model fitting. Typically,  N-fold cross-validation could be used to split train samples and validation samples for the calibration. The probabilities predicted for the folds are then averaged.

core/src/main/java/smile/classification/IsotonicRegressionScaling.java
```diff
@@ -0,0 +1,178 @@
...
+            throw new IllegalArgumentException("fail to get the posteriori probability for given prediction: " + y);
+        }
+        return ret;
+    }
+
+    /**
+     * Estimates the multiclass probabilies.
+     */
+    public static void multiclass(int k, double[][] r, double[] p) {
+        PlattScaling.multiclass(k, r, p);
```
haifengl: Can we do this? Do you have a reference?

core/src/main/java/smile/classification/SVM.java
```diff
@@ -1533,15 +1698,19 @@ private double posterior(LASVM svm, double y) {
         final double minProb = 1e-7;
         final double maxProb = 1 - minProb;
 
-        return Math.min(Math.max(svm.platt.predict(y), minProb), maxProb);
+        return Math.min(Math.max(scalingMethod == ScalingMethod.Platt? svm.platt.predict(y) : svm.isotonicRegression.predict(y), minProb), maxProb);
```
haifengl: no sure about this. simple prediction without postiori probability should not bother this at all.

core/src/main/java/smile/classification/IsotonicRegressionScaling.java
rayeaster: No. Reference for platt does not mention multi-class calculation neither: https://www.researchgate.net/profile/John_Platt/publication/2594015_Probabilistic_Outputs_for_Support_Vector_Machines_and_Comparisons_to_Regularized_Likelihood_Methods/links/004635154cff5262d6000000.pdf

core/src/main/java/smile/classification/SVM.java
rayeaster: this is just a switch added on top of existing logic. If simple prediction without postiori probability does not need concern about scaling, why do we need existing platt scaling in this method?

rayeaster:
add friendly `toString` to output pair-wise lines generated by Isotonic Regression for better illustration of the scaling effect. Also add straightforward assertion check to ensure we cover all prediction in fitted lines.

core/src/main/java/smile/classification/IsotonicRegressionScaling.java
```diff
@@ -0,0 +1,220 @@
...
+    private double trainMaxScore = Double.NaN;
+
+    /** setep-wise constant functions */
+    private static class StepwiseConsFunc {
+        double lowScore;
+        double highScore;
+        double val;
+        int weight = 1;
+
+        StepwiseConsFunc next;
```
haifengl: Java has LinkedList. We don't need to reinvent the wheel. In fact, we don't need linked list at all.

core/src/main/java/smile/classification/IsotonicRegressionScaling.java
```diff
@@ -0,0 +1,220 @@
...
+        StepwiseConsFunc previous;
+
+        StepwiseConsFunc(double lowScore, double highScore, double val) {
+            this.lowScore = lowScore;
+            this.highScore = highScore;
+            this.val = val;
+        }
+
+        void setNext(StepwiseConsFunc next) {
```
haifengl: It is not necessary to use getter and setter for inner class. They make codes cluttered and don't really bring in benefits of encapsulations. You are the only user of this class!

core/src/main/java/smile/classification/IsotonicRegressionScaling.java
```diff
@@ -0,0 +1,220 @@
...
+     * @param y
+     *            The training labels.
+     */
+    public IsotonicRegressionScaling(double[] scores, int[] y) {
+        initStepwiseConstFuncs(scores, y);
+
+        StepwiseConsFunc current = startConstantFuncs.next;
+        StepwiseConsFunc last = startConstantFuncs;
+        
+        while (current != null) {
```
haifengl: We need actually check the PAV in multiple iterations. Think about it as an exercise.

core/src/main/java/smile/classification/IsotonicRegressionScaling.java
```diff
@@ -0,0 +1,220 @@
...
+ * <li>Niculescu-Mizil, Alexandru; Caruana, Rich: Predicting Good Probabilities
+ * With Supervised Learning, 2005.</li>
+ * </ol>
+ *
+ * @author rayeaster
+ */
+public class IsotonicRegressionScaling implements Serializable {
+    private static final long serialVersionUID = 1L;
+    
+    private double trainMixScore = Double.NaN;
```
haifengl: They are working variables. Should not be member variables.

core/src/main/java/smile/classification/IsotonicRegressionScaling.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /**
+     * Returns the posterior probability estimate P(y = 1 | x).
+     *
+     * @param y
+     *            the binary classifier output score.
+     * @return the estimated probability.
+     */
+    public double predict(double y) {
```
haifengl: This is really slow. It is linear. It can be done in O(log n).

haifengl:
I implement it in v2 now. Thanks!

haifengl:
check out https://github.com/haifengl/smile/blob/v2/core/src/main/java/smile/classification/IsotonicRegressionScaling.java

rayeaster:
I will close this PR

PR #357, opened 2018-09-07
Bugfix Dataset.iterator()

gforman44:
The iterator remove() operation used the index relative to the original dataset, but if you've already removed some items, then it removes the wrong positions.


haifengl:
Thanks. I cherry picked your fix.

PR #359, opened 2018-09-14
remove unnecessary code in decision/regression tree build

rayeaster:
remove unnecessary lines of code which will not be executed under any circumstance. Since `nextSplits` queue is passed in the construstor and always be initialized. so `null` check could be skipped safely.


rayeaster:
@haifengl I have several questions about the `RegressionTree `implementation as following:

1) this code snippet seems not a typical squared error calculation and in which the `split.output` actually never get updated (means it is always zero)
`(trueCount * trueMean^2 + falseCount * falseMean^2) - n * split.output^2`

2) why do we use average instead the `x[i][j]` directly as split value? but we indeed count the `samples[i]` already as true child: `split.splitValue = (x[i][j] + prevx) / 2` . since average will make the actual split value is lower than the value of `x[i][j]` if I understand correctly

3) sorry would you please share the reference for monotonic constraint? It is a little hard to understand the trick behind the implementation: `score *= 1 - Math.abs(monoRegForFeature)`

Thanks!

myui:
@haifengl 

Related to this topic, why did you removed [depth-first split](https://datascience.stackexchange.com/questions/26699/decision-trees-leaf-wise-best-first-and-level-wise-tree-traverse) in [this commit](https://github.com/haifengl/smile/commit/d3e4afc4168aa0f8f47c292103fb233896cf6e45#diff-aa1c5dce2c959f46efdcb90067706154L998) and changed splitting scheme only to best-first split?

`trainRoot.split(null);` was used for depth-first split in the old rev of smile.

I'm not sure which is good but [Scikit uses depth-first split](https://github.com/scikit-learn/scikit-learn/blob/2fc3a85b3fa298c57728e657e6a63dedaa1935b3/sklearn/tree/tree.py#L363) by the default.  In general, the default `maxNodes = 100` is too small for certain conditions and tuning `max_leaf_nodes` beforehand is hard and thus depth-first is used.

I think it's better to use depth-first split for most cases [as it was](https://github.com/apache/incubator-hivemall/blob/master/core/src/main/java/hivemall/smile/classification/DecisionTree.java#L1040).

core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -711,20 +711,12 @@ public boolean split(PriorityQueue<TrainNode> nextSplits) {
             
             TrainNode trueChild = new TrainNode(node.trueChild, x, y, trueSamples);
             if (tc > nodeSize && trueChild.findBestSplit()) {
-                if (nextSplits != null) {
-                    nextSplits.add(trueChild);
-                } else {
-                    trueChild.split(null);
-                }
+                nextSplits.add(trueChild);
```
haifengl: The original code is to support both best-first split and deep-first split.

core/src/main/java/smile/regression/RegressionTree.java
```diff
@@ -548,7 +548,7 @@ public Node findBestSplit(int n, double sum, int j) {
                     double trueMean = trueSum[k] / tc;
                     double falseMean = (sum - trueSum[k]) / fc;
 
-                    double gain = (tc * trueMean * trueMean + fc * falseMean * falseMean) - n * split.output * split.output;
+                    double gain = (tc * trueMean * trueMean + fc * falseMean * falseMean);
```
haifengl: gain is about impurity reduction. This is wrong change.

haifengl:
We have completely rewrite the CART related code. Please check v2 branch.

PR #360, opened 2018-09-23
add ReducedErrorPruning for decision/regression tree

rayeaster:
https://en.wikipedia.org/wiki/Pruning_(decision_trees)

- add increasing index to each node to distinguish each other;
- add pointer to node's parent;
- add arrays to record response and prediction respectively when passing validation dataset through trained tree for later use in pruning
- test result shows the model has better accuracy for testing data after pruning:

```
classification:
USPS error rate = 20.23%
pruning node #69396
......
After pruning, USPS error rate = 17.99%


regression:
Puma RMSE = 4.0809
pruning node #280
......
After pruning Puma RMSE = 4.0411
```

### DIFF
--- a/core/src/main/java/smile/classification/DecisionTree.java
+++ b/core/src/main/java/smile/classification/DecisionTree.java
@@ -461,7 +461,7 @@ public class DecisionTree implements SoftClassifier<double[]> {
 			// start post pruning
 			Node candidate = candidates.size() == 0 ? null : candidates.remove(candidates.size() - 1);
 			while (candidate != null) {
-				boolean prune = pruneSubTree(candidate, candidates, flags);
+				boolean prune = pruneSubTree(dt, candidate, candidates, flags);
 				if (prune) {
 					candidates.sort(c);
 				}
@@ -497,7 +497,7 @@ public class DecisionTree implements SoftClassifier<double[]> {
 			return new int[] { error, total };
 		}
 
-		private static boolean pruneSubTree(Node n, List<Node> candidates, Set<Integer> flags) {
+		private static boolean pruneSubTree(DecisionTree dt, Node n, List<Node> candidates, Set<Integer> flags) {
 			boolean ret = false;
 
 			Map<Integer, Integer> respCount = new HashMap<Integer, Integer>();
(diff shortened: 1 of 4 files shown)


haifengl:
DecisionTree has been fully rewriten in v2. Thanks!

rayeaster:
> DecisionTree has been fully rewriten in v2. Thanks!

Looking forward to the release of v2 :) Really appreciate making smile such wonder tool! 
@haifengl

haifengl:
Do you have a reference to this? There are so many variants of reduced error pruning. Thanks

rayeaster:
> Do you have a reference to this? There are so many variants of reduced error pruning. Thanks

There is a descriptive reference (no pesuedo code) for this implementation at [http://www-scf.usc.edu/~csci567/21-penalty-methods.pdf](http://www-scf.usc.edu/~csci567/21-penalty-methods.pdf).  although I simply use naive absolute-value comparison between error rate before & after pruning.

IMO wiki decription is quite straight-forward:
```
Starting at the leaves
each node is replaced with its most popular class
If the prediction accuracy is not affected then the change is kept.
```

haifengl:
Thanks. This article https://arxiv.org/pdf/1106.0668.pdf gives an analysis in depth.

rayeaster:
> Thanks. This article https://arxiv.org/pdf/1106.0668.pdf gives an analysis in depth.

paste the pesudo code from above reference here:

image

haifengl:
People disagree on the last 4 lines. In this pseudo code, the merge node prediction label will be based on test samples. Other people prefer keeping the label of training samples.

rayeaster:
> People disagree on the last 4 lines. In this pseudo code, the merge node prediction label will be based on test samples. Other people prefer keeping the label of training samples.

classification implementatiton in this PR follows with the pesuedo code by using test sample labels. If you prefer the other way or want to add a parameter for this situation, please let me know. @haifengl

haifengl:
i have implemented this in a concise way. It is in v2 branch.

rayeaster:
> i have implemented this in a concise way. It is in v2 branch.

sounds good. I will close this PR. Thanks

PR #366, opened 2018-10-09
Max iteration for smo in SVM

digital-thinking:
Adding a maximum for iterations, when convergence criteria is not enough.

### DIFF
--- a/core/src/main/java/smile/classification/SVM.java
+++ b/core/src/main/java/smile/classification/SVM.java
@@ -142,6 +142,10 @@ public class SVM <T> implements OnlineClassifier<T>, SoftClassifier<T>, Serializ
      * The tolerance of convergence test.
      */
     private double tol = 1E-3;
+    /**
+     * The max iteration count of convergence test.
+     */
+    private int maxIter = Integer.MAX_VALUE;
     
     /**
      * Trainer for support vector machines.
@@ -180,6 +184,10 @@ public class SVM <T> implements OnlineClassifier<T>, SoftClassifier<T>, Serializ
          * The number of epochs of stochastic learning.
          */
         private int epochs = 2;
+        /**
+         * The max iteration count of convergence test.
+         */
+        private int maxIter = Integer.MAX_VALUE;
 
         /**
          * Constructor of trainer for binary SVMs.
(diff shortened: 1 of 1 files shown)


rayeaster:
It would be better to separate the change from MCC metric change and do not use code format for existing code:)

digital-thinking:
Ohh yes, sorry I updated the pr

haifengl:
Why do we need this? First of all, this doesn't control the iterations of SVM. Second, for `finish()`, we need to reprocess all SVs to refine the model. If this part is too long to you, it means that there are too many SVs and you should tuning your kernel, hyper-parameters, regularization, etc.

digital-thinking:
It happens, that the training gets stuck in an infinite loop, when the tolerance is never achieved, see the smo call in the loop (line 866). I am working on model selection and I am tuning hyper parameters programmatically, so if a SVM gets stuck, i need some limit to make sure that the training stops at some point.

haifengl:
If `smo` doesn't return, we better figure out why. Your revision doesn't solve the real problem. Can you share your data and settings that causes it never returns?

haifengl:
You actually can pass a parameter `epsgr` to `finish()` to reduce the iterations. Why don't you try that?

digital-thinking:
I will try to provide some data, but in general I can not determine how to set epsgr before finish() is invoked. It depends on the data, sometimes it´s necessary to set epsgr to 0.01 0.1 or even 0.2. If epsgr is to low, it will get stuck and never reach the criteria (infinite loop). When this happens I can adjust the epsgr param manually, but in grid search/model selection I have to restart the process.

haifengl:
I add `maxIter` to `finish()`. If `reprocess` doesn't finish timely, it is more likely that your kernel and hyper-parameters are not suitable. I would rather to try other algorithms.

haifengl:
`CrossValidation` is only for generate the random splits (index). You can call the `cv` function (in Scala), which take a code block that you can do anything you want. BTW, you can call a scala function in Java too.

digital-thinking:
In my case I am using static `Validation.cv()`, which accepts  (number of folds, the trainer, X, Y and measures). Even If I implement this on my own, I can not pass the maxIter param, because in the end `finish()` is called inside of the `train()` Method from the SVM and I don´t have any point, where I could call `finish(int maxIter)`.

haifengl:
For now, you can use your own branch build. I will revise the interface in next release.

digital-thinking:
Would be nice, thank you!

PR #369, opened 2018-10-18
master

digital-thinking:
FeatureTransform serializable


digital-thinking:
I decided to do so after looking at the childs classes (should work for all current extended classes). Would be also an option to just make the Child classes Serializable.
I found helpful [this article](https://www.geeksforgeeks.org/object-serialization-inheritance-java/)

haifengl:
Cool. Thanks.

Issue #377, opened 2018-11-09
[Feature Proposal] GroupKFold cross validation


inejc:
I'm interested in implementing the `GroupKFold` cross validation technique, similar to [scikit-learn's implementation](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GroupKFold.html). The procedure is useful when the i.i.d. assumption is known to be broken by the process generating the data, e.g. when we have multiple data points by the same user and want to make sure that the model learns the concept we are interested in rather than user-specific features that don't generalize to unseen users.

Example of usage:
```java
int n = 10;
int k = 3;
int[] groups = new int[] {1, 2, 2, 0, 0, 0, 2, 1, 1, 2};

GroupKFold split = new GroupKFold(n, k, groups);

for (int i = 0; i < k; i++) {
    int[] trainIndices = split.train[i];
    int[] testIndices = split.test[i];
}
```

inejc:
I went ahead and implemented the above as I needed it in my work.

haifengl:
Thanks

PR #378, opened 2018-11-09
Implement the GroupKFold functionality

inejc:
Implements the proposed feature in https://github.com/haifengl/smile/issues/377.


haifengl:
Thanks!

PR #399, opened 2019-01-07
Fix bug in Heatmap.java: x and y arrays are not passed to constructor

koendeschacht:
By the way, thanks for all the great work on this project!


haifengl:
Thanks!

PR #409, opened 2019-03-11
Refactor cv, add shuffling

cfga:
Added shuffling parameter for cv function -- this is the primary goal of the pr.

Also refactored parameter lists, so that first comes the data, then the model, and in the end the metrics.

### DIFF
--- a/scala/src/main/scala/smile/validation/Operators.scala
+++ b/scala/src/main/scala/smile/validation/Operators.scala
@@ -334,6 +334,7 @@ trait Operators {
     * @param k k-fold cross validation.
     * @param measures validation measures such as accuracy, specificity, etc.
     * @param trainer a code block to return a classifier trained on the given data.
+    * @param shuffle whether the data should be shuffled before running cross validation.
     * @return measure results.
     */
   def cv[T <: Object](x: Array[T], y: Array[Int], k: Int, shuffle: Boolean = false)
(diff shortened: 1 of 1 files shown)


haifengl:
Thanks. But it is important to keep trainer as the last parameter so that we can write a code block to train the model. For example,

```
cv(x, y, 10) { (x, y) => randomForest(x,y) }
```

cfga:
@haifengl that makes sense. What if I just swap measures and trainer?

haifengl:
Your goal is to add `shuffle` parameter. Why do you change the signature of `loocv` and `bootstrap`? `loocv` has no shuffle and `bootstrap` must have shuffle implicitly. I understand that sometimes you don't want shuffle for cv (mostly to get same result in test). I don't understand the benefit of putting the parameters `measures`  and `trainer` in additional parentheses. 

BTW, the default value of `shuffle` should be true. It is recommended to shuffle the samples for cv.

haifengl:
To resolve the conflict between the default value of `shuffle` and varargs `measures`, I keep the existing signatures and add new methods that take a parameter of `CrossValidation` instead of `k: Int`. So you can pass a customized `CrossValidation` with `shuffle = false`.

This should be a special case, only for test purpose. We also keep the old simple interface. I believe that it is a good balance.

haifengl:
BTW, it is in the master branch now. Check it out.

PR #414, opened 2019-03-21
bugfix in Math.java

gforman44:
there was a bug in c(int[]... x) at line 930 (and other similar).  x.length was the # arguments to the function, not the length of one of the vectors.  Fixed.  Also, the c(String[]... x) can take a generic type and use .toString() to collect the results.  If already a String, it's a no-op.

Also, I think cat (as in catenate) would be a better name than just c, if you're willing to change it.

### DIFF
--- a/math/src/main/java/smile/math/Math.java
+++ b/math/src/main/java/smile/math/Math.java
@@ -924,64 +924,68 @@ public class Math {
     }
 
     /** Merges multiple vectors into one. */
-    public static int[] c(int[]... x) {
+    public static int[] c(int[]... list) {
         int n = 0;
-        for (int i = 0; i < x.length; i++) {
+        for (int[] x: list) {
             n += x.length;
         }
 
         int[] y = new int[n];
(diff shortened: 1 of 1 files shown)


math/src/main/java/smile/math/Math.java
```diff
@@ -924,64 +924,68 @@ public static void permutate(Object[] x) {
...
+            for (double xi: x) {
+                y[i++] = xi;
             }
         }
         return y;
     }
 
-    /** Merges multiple vectors into one. */
-    public static String[] c(String[]... x) {
+    /** Concatenates multiple vectors into one array of strings. Uses .toString(), which is a no-op if already strings. */
```
haifengl: It is unlikely to merge multiple arrays to Strings. No need to make this function generic.

haifengl:
Thanks. Let's keep it simple (just for String[], not T[]).

haifengl:
I made some optimization in the implementation. It is in master now.

PR #427, opened 2019-04-30
Minor Typo in FastPair Method updateDistance

jonnyboyC:
It appears that the distance array was incorrectly set the point index instead of the calculate linkage distance. Based on other usage of linkage distance in this class this appears to be the correct behavior.

### DIFF
--- a/core/src/main/java/smile/clustering/FastPair.java
+++ b/core/src/main/java/smile/clustering/FastPair.java
@@ -217,14 +217,14 @@ class FastPair {
         float d = linkage.d(p, q);
 
         if (d < distance[p]) {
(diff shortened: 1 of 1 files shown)


haifengl:
Thanks. Have you run all unit tests?

jonnyboyC:
I've attempted to run all the unit tests, it doesn't appear I have my environment setup correctly. Many of the test fail with file not found exceptions. The HierarchicalClusteringTest does pass. If it would be easier, I can close this pr and post an issue describing problem.

haifengl:
No worries. Let me merge it and test by myself. Thanks.

PR #440, opened 2019-07-10
[Proposal][Discuss] Accept general Matrix in train/predict

myui:
How about accepting Matrix for train/predict?

I was willing to contribute [SparseMatrix supports in DecisionTree in addition to DenseMatrix](https://github.com/apache/incubator-hivemall/blob/master/core/src/main/java/hivemall/smile/classification/DecisionTree.java#L962
) but it introduces a lot of changes in interface and thus resigned to make a PR.

It would be better to discuss API design in advance and make a consensus. Here is a concept patch that works with the existing implementations.

Ideally, all algorithm overrides `Classifier<T> train(Matrix x, int[] y)` in the future.

Scikit support dense array as well as CSR sparse matrix for [fitting](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html#sklearn.linear_model.SGDClassifier.fit).

### DIFF
--- a/core/src/main/java/smile/classification/Classifier.java
+++ b/core/src/main/java/smile/classification/Classifier.java
@@ -16,6 +16,10 @@
 
 package smile.classification;
 
+import smile.math.matrix.DenseMatrix;
+import smile.math.matrix.Matrix;
+import smile.math.matrix.SparseMatrix;
+
 import java.io.Serializable;
 
 /**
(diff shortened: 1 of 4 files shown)


haifengl:
In v2, we take DataFrame, which is more flexible than Matrix.

myui:
I guess it's a hard task to rewrite existing implementations but I see.

In my understanding, Dataframe is a rich data structure (in terms of memory space) mostly for EDA and feature engineering.

For training (e.g., sgd), algorithms are required to aware underlying matrix (e.g., [csr or not](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/tree/_tree.pyx#L780)) for efficient implementation. And then, Dataframe's abstraction is too much to achieve such kind of optimization.

In Pandas dataframe, [to_numpy](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_numpy.html#pandas.DataFrame.to_numpy) or [to_coo](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.SparseDataFrame.to_coo.html#pandas.SparseDataFrame.to_coo) is used to feed features in double[][] or Matrix. And, I think it's enough for classifier/regressors.

haifengl:
Not all algorithms will take DataFrame as input.

Issue #438, opened 2019-07-09
Reduce memory consumption of DBSCAN


vlieven:
Hi,

First of all, thank you for the work on this library. While using the DBSCAN algorithm, I noticed that the memory consumption is rather high. After looking at the source code, I believe this can be mitigated by expanding the neighbors in a depth-first way, instead of breadth-first.

If you are open to it, I can make a pull request with the proposed changes?

haifengl:
It is very easy to hit stack issue with DFS in DBScan. The memory consumption depends on the nearest neighbor search data structure, and DBScan parameters. If it consumes too much memory, it might be caused by large radius parameter.

maciejbrzeski-compactsolutions:
I have a similar problem, but there is another, in my opinion, more seriously, problem. There is no checking that some point is in the queue, what causes memory is quadratic of the currently processed cluster.

haifengl:
@maciejbrzeski-compactsolutions it may be a cause of high memory usage. I will look into it. Thanks.

vlieven:
I've created a pull request with the changes I made to solve the memory issue I encountered when processing my dataset. It should also solve the issue of @maciejbrzeski-compactsolutions I believe. Feel free to have a look and let me know if any changes are required.

haifengl:
The improvement is v2 branch now.

PR #441, opened 2019-07-11
Rework DBSCAN to reduce memory consumption

vlieven:
As indicated in issue #438 the memory consumption of the DBSCAN algorithm can be reduced. This change modifies the algorithm to use an ArrayDeque instead of ArrayList to hold the collection of neighbors. A BitSet is used to keep track of which observations have already been visited.


haifengl:
Thanks. I make a small change to check if a neighbor is already in the queue. No use of additional data structure. It is in v2 branch.

vlieven:
Thanks for the change!

PR #443, opened 2019-07-18
PartitionClustering should be Serializable

pguillebert:
Hi

I'm trying to serialize a DBSCAN instance, store it on disk, and deserialize when needed in order to use the `predict` method. the goal is to evaluate new datapoints without recomputing everything.

The issue is that even if `DBSCAN<T>` is declared Serializable, its parent class PartitionClustering is not. So java serialization process will NOT serialize fields inherited from the parent class (`k`, `y`, `size`). That makes the deserialized class unusable since these values are lost.

This PR just add Serializable to the PartitionClustering class.


haifengl:
Thanks

Issue #446, opened 2019-08-01
Need some iterators on SparseMatrix


tdunning:
## Expected behaviour

I have a need to be able to iterate over nonzero elements of a sparse matrix, but retaining information about which element I am seeing. Ideally, I would also like enough information so that I can do fast in-place updates of elements in the matrix. Also ideally, I would like a very flexible API so that I can do things like filter which elements I operate on and accumulate results in flexible ways.

Related to this is a need to accumulate data in row,column,value format preparatory to reformatting the data as an actual sparse matrix.

Examples of where this is useful include:

- scanning a column to find the 100 biggest elements

- evaluating column-wise functions such as scaled sums

- performing complex element-wise operations, possibly with reference to other data structures beyond just the element in-hand and with or without the result having different sparsity structure from the original data.

- accumulating count data into a sparse matrix

## Actual behaviour
I propose (and have implemented prototypes of) one new class, `CooData` and four new methods on SparseMatrix:
```java
    Stream<MatrixEntry> nonzeros() 
    Stream<MatrixEntry> nonzeros(int startColumn, int endColumn)
    void foreachNonzero(int startColumn, int endColumn, MatrixElementConsumer action)
    void foreachNonzero(MatrixElementConsumer action)
```
The `nonzero` methods return a stream so that I can use the full flexibility of the Stream idiom for operations such as collecting values into a Collection. The `foreachNonZero` methods can be faster because they involve less boxing into a `MatrixEntry` structure, but are considerably less flexible. I have found both important for work in natural language processing, particularly term cooccurrence analysis.
## Code snippet
Here is some code that transforms a stream of documents, represented as `String` into a sparse matrix containing the per-document word counts.
```java
        CooData binaryTerms = docs(p, nDocs)
                .collect(
                        CooData::new,
                        (CooData m, String raw) -> {
                            int currentDoc = docid.getAndIncrement();
                            // downsample our words according to limit max frequency
                            // and translate to integer form
                            CooData words = VectorText.tokenize(raw)
                                    .filter(w -> dict.containsKey(w) && (rand.nextDouble() < frequencyCut / counts.count(w)))
                                    .collect(
                                            () -> m,
                                            (CooData mx, String w) -> mx.add(currentDoc, dict.get(w), 1.0),
                                            CooData::append);
                        },
                        CooData::append);
        SparseMatrix docByTerms = binaryTerms.asSparseMatrix();
```
[...]

PR #447, opened 2019-08-05
Add functional iterators to SparseMatrix

tdunning:
This relates to issue #446 (Need some iterators on SparseMatrix)

### DIFF
--- a/math/src/main/java/smile/math/matrix/SparseMatrix.java
+++ b/math/src/main/java/smile/math/matrix/SparseMatrix.java
@@ -41,6 +41,39 @@ import smile.math.Math;
  * of pointers are indexes where each column starts. This format is efficient
  * for arithmetic operations, column slicing, and matrix-vector products.
  * One typically uses SparseDataset for construction of SparseMatrix.
+ * <p>
+ * For iteration through the elements of a matrix, this class provides
+ * a functional API to iterate through the non-zero elements. This iteration
+ * can be done by passing a lambda to be called on each non-zero element or
+ * by processing a stream of objects representing each non-zero element.
+ * The direct functional API is faster (and is about as fast as writing the
+ * low-level loops against the internals of the matrix itself) while the
+ * streaming interface is more flexible. Here are some benchmarks that were
+ * produced using jmh (and allowing access to internal data structures):
+ * <verbatim>
+ * Benchmark                                       Mode  Cnt        Score       Error   Units
+ * IteratorSpeed.timeDirect                        avgt    5   429888.246 ±  3819.232   ns/op
+ * IteratorSpeed.timeDirect:·gc.alloc.rate         avgt    5       ≈ 10⁻⁴              MB/sec
+ * IteratorSpeed.timeDirect:·gc.alloc.rate.norm    avgt    5        0.088 ±     0.001    B/op
+ * IteratorSpeed.timeDirect:·gc.count              avgt    5          ≈ 0              counts
+ * IteratorSpeed.timeDirect:·stack                 avgt               NaN                 ---
+ * IteratorSpeed.timeIterator                      avgt    5   430718.537 ±  7831.509   ns/op
+ * IteratorSpeed.timeIterator:·gc.alloc.rate       avgt    5        0.028 ±     0.001  MB/sec
+ * IteratorSpeed.timeIterator:·gc.alloc.rate.norm  avgt    5       16.089 ±     0.011    B/op
+ * IteratorSpeed.timeIterator:·gc.count            avgt    5          ≈ 0              counts
+ * IteratorSpeed.timeIterator:·stack               avgt               NaN                 ---
+ * IteratorSpeed.timeStream                        avgt    5  1032370.658 ± 55295.704   ns/op
+ * IteratorSpeed.timeStream:·gc.alloc.rate         avgt    5        0.077 ±     0.004  MB/sec
+ * IteratorSpeed.timeStream:·gc.alloc.rate.norm    avgt    5      104.210 ±     0.011    B/op
+ * IteratorSpeed.timeStream:·gc.count              avgt    5          ≈ 0              counts
+ * IteratorSpeed.timeStream:·stack                 avgt               NaN                 ---
+ * </verbatim>
+ * The three cases are timeDirect for a direct loop over internal data structures,
+ * timeIterator for #foreachNonZero and timeStream for the streaming equivalent form. The
+ * timeIterator case is at most a few percent slower than the direct loops while the stream
+ * is about 2-3 times slower. Note that the JVM is clever enough to optimize away the
+ * creation of temporary objects in the streaming idiom.
+ * </p>
  *
  * @author Haifeng Li
  */
@@ -174,29 +207,6 @@ public class SparseMatrix implements Matrix, MatrixMultiplication<SparseMatrix, 
     }
 
     /**
-     * Returns all nonzero values.
-     * @return all nonzero values
-     */
-    public double[] values() {
-        return x;
-    }
-
-    /**
-     * Returns the row index
-     */
-    public int[] getRowIndex() {
-        return rowIndex;
-    }
-
-    /**
-     * Returns the column index
-     */
-    @SuppressWarnings("WeakerAccess")
-    public int[] getColIndex() {
-        return colIndex;
-    }
-
-    /**
      * Calls a lambda with each non-zero value. This will be a bit faster than using a stream
      * because we can avoid boxing up the coordinates of the element being processed, but it
      * will be considerably less general.
@@ -212,7 +222,7 @@ public class SparseMatrix implements Matrix, MatrixMultiplication<SparseMatrix, 
         for (int j = 0; j < ncols; j++) {
             for (int k = colIndex[j]; k < colIndex[j + 1]; k++) {
                 int i = rowIndex[k];
-                action.eval(i, j, x[k]);
+                action.accept(i, j, x[k]);
             }
         }
     }
@@ -229,10 +239,10 @@ public class SparseMatrix implements Matrix, MatrixMultiplication<SparseMatrix, 
      *
      * @param startColumn The first column to scan.
      * @param endColumn   One past the last column to scan.
-     * @param action      Action to perform on each non-zero, typically this is a lambda.
+     * @param consumer      Action to perform on each non-zero, typically this is a lambda.
      */
     @SuppressWarnings("WeakerAccess")
-    public void foreachNonzero(int startColumn, int endColumn, MatrixElementConsumer action) {
+    public void foreachNonzero(int startColumn, int endColumn, MatrixElementConsumer consumer) {
         if (startColumn < 0 || startColumn >= ncols) {
             throw new IllegalArgumentException("Start column must be in range [0,ncols)");
         }
@@ -243,7 +253,7 @@ public class SparseMatrix implements Matrix, MatrixMultiplication<SparseMatrix, 
         for (int j = startColumn; j < endColumn; j++) {
             for (int k = colIndex[j]; k < colIndex[j + 1]; k++) {
                 int i = rowIndex[k];
-                action.eval(i, j, x[k]);
+                consumer.accept(i, j, x[k]);
             }
         }
     }
@@ -252,7 +262,7 @@ public class SparseMatrix implements Matrix, MatrixMultiplication<SparseMatrix, 
      * Provides a stream over all of the non-zero elements of a sparse matrix.
      */
     @SuppressWarnings("WeakerAccess")
-    public Stream<SparseMatrixEntry> nonzeros() {
+    public Stream<Entry> nonzeros() {
         return StreamSupport.stream(new SparseMatrixSpliterator(this, 0, ncols), false);
     }
 
@@ -263,7 +273,7 @@ public class SparseMatrix implements Matrix, MatrixMultiplication<SparseMatrix, 
      * @param endColumn   The first column after the ones that we should scan.
      */
     @SuppressWarnings("WeakerAccess")
-    public Stream<SparseMatrixEntry> nonzeros(int startColumn, int endColumn) {
+    public Stream<Entry> nonzeros(int startColumn, int endColumn) {
         return StreamSupport.stream(new SparseMatrixSpliterator(this, startColumn, endColumn), false);
     }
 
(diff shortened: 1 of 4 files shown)


math/src/main/java/smile/math/matrix/SparseMatrix.java
```diff
@@ -175,11 +180,204 @@ public int size() {
...
+        public SparseMatrixEntry clone(double value) {
+            return new SparseMatrixEntry(row, col, value, index, values);
+        }
+    }
+
+    /**
+     * Functional interface for lambda iteration through a sparse matrix.
+     */
+    public interface MatrixElementConsumer {
+        void eval(int row, int column, double value);
```
haifengl: `accept` is typically used as the method name of a consumer.

math/src/main/java/smile/math/matrix/SparseMatrix.java
```diff
@@ -175,11 +180,204 @@ public int size() {
...
+
+        public SparseMatrixEntry clone(double value) {
+            return new SparseMatrixEntry(row, col, value, index, values);
+        }
+    }
+
+    /**
+     * Functional interface for lambda iteration through a sparse matrix.
+     */
+    public interface MatrixElementConsumer {
```
haifengl: This is a generic interface. It is better not as an inner class.

math/src/main/java/smile/math/matrix/SparseMatrix.java
```diff
@@ -175,11 +180,204 @@ public int size() {
...
+        @SuppressWarnings("WeakerAccess")
+        public SparseMatrixEntry(int row, int col, double x, int index, double[] values) {
+            this.row = row;
+            this.col = col;
+            this.x = x;
+            this.index = index;
+            this.values = values;
+        }
+
+        public void set(double value) {
```
haifengl: This breaks our contract that `SparseMatrix` is immutable.

math/src/main/java/smile/math/matrix/SparseMatrix.java
```diff
@@ -175,11 +180,204 @@ public int size() {
...
+        return StreamSupport.stream(new SparseMatrixSpliterator(this, 0, ncols), false);
+    }
+
+    /**
+     * Provides a stream over all of the non-zero elements of range of columns of a sparse matrix.
+     *
+     * @param startColumn The first column to scan
+     * @param endColumn   The first column after the ones that we should scan.
+     */
+    @SuppressWarnings("WeakerAccess")
```
haifengl: I am really concerned that there are so many SuppressWarnings, which indicates some design flaws.

math/src/main/java/smile/math/matrix/SparseMatrix.java
```diff
@@ -175,11 +180,204 @@ public int size() {
...
+                action.eval(i, j, x[k]);
+            }
+        }
+    }
+
+    /**
+     * Provides a stream over all of the non-zero elements of a sparse matrix.
+     */
+    @SuppressWarnings("WeakerAccess")
+    public Stream<SparseMatrixEntry> nonzeros() {
```
haifengl: Shall we just call it `stream`? It is more consistent with overall Java Stream design.

math/src/main/java/smile/math/matrix/SparseMatrix.java
```diff
@@ -175,11 +180,204 @@ public int size() {
...
+     * will be considerably less general.
+     * <p>
+     * Note that the action could be called on values that are either effectively or actually
+     * zero. The only guarantee is that no values that are known to be zero based on the
+     * structure of the matrix will be processed.
+     *
+     * @param action Action to perform on each non-zero, typically this is a lambda.
+     */
+    @SuppressWarnings("WeakerAccess")
+    public void foreachNonzero(MatrixElementConsumer action) {
```
haifengl: How about just `foreach` as the method name?

math/src/main/java/smile/math/matrix/SparseMatrix.java
```diff
@@ -175,11 +180,204 @@ public int size() {
...
+     * so that in-place updates are possible.
+     */
+    public static class SparseMatrixEntry {
+        // these fields are exposed for direct access to simplify in-lining by the JVM
+        public final int row;
+        public final int col;
+        public double x;
+
+        // these are hidden due to internal dependency
+        private final int index;
```
haifengl: They should be removed when we remove `set` method.

math/src/main/java/smile/math/matrix/SparseMatrix.java
```diff
@@ -175,11 +180,204 @@ public int size() {
...
+            return ORDERED | SIZED | IMMUTABLE | SUBSIZED;
+        }
+    }
+
+    /**
+     * Encapsulates important information about an entry in a matrix for use
+     * in streaming, including a string that leads back to the original cell
+     * so that in-place updates are possible.
+     */
+    public static class SparseMatrixEntry {
```
haifengl: Since it is an inner class, we can just call it `Entry`.

math/src/main/java/smile/math/matrix/SparseMatrix.java
```diff
@@ -175,11 +180,204 @@ public int size() {
     public double[] values() {
         return x;
     }
-    
+
```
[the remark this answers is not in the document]
tdunning: It is fine to hide these, but then it becomes difficult to demonstrate the speed without putting test code into the SparseMatrix implementation.

If we don't need the speed test, then these can be dropped.

I should point out that access to the values was already there before my contribution so you may cause problems deleting it.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: Fine.

The name almost never appears in user code anyway.

A consumer, however, normally only takes a single argument, I think.

Perhaps a better approach is to to change the name from Consumer to Action or Function.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: Fine.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: The contract was already broken. And was undocumented.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: It just indicates that the only user of these functions is a test case that is in the same package while the ultimate user of the function is in user code.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: That would also be misleading since it isn't a stream over the entire matrix. Precedent for a specialized name for a stream can be found in [Files.lines()](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#lines-java.nio.file.Path-)

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: Again, this would be misleading. It doesn't iterate over all elements of the matrix. It only iterates over non-zeros.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: The set method shouldn't be removed.

math/src/main/java/smile/math/matrix/SparseMatrix.java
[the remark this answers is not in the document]
haifengl: I will remove `data()` method. Let's comment out the speed test for now. Can you include speed test results in the comments?

math/src/main/java/smile/math/matrix/SparseMatrix.java
haifengl: `Function` returns values while `Consumer` not. `Action` is mostly used in GUI things. There are `BiConsumer` taking two arguments. I think that it is fine to call it `Consumer`. `accept` is better name than `eval`. We are not really evaluate the element value here.

math/src/main/java/smile/math/matrix/SparseMatrix.java
haifengl: In v2, we enhance the Matrix interface. Only `DenseMatrix` is mutable.

math/src/main/java/smile/math/matrix/SparseMatrix.java
haifengl: Can you elaborate?

math/src/main/java/smile/math/matrix/SparseMatrix.java
haifengl: Understood. We prefer short method name and the documents state clearly that it iterates only nonzero values. We don't need make code as verbose as Objective-C.

math/src/main/java/smile/math/matrix/SparseMatrix.java
haifengl: We want to keep SparseMatrix immutable. SparseMatrix is mostly used for matrix operations such as matrix-vector multiplication and matrix-matrix multiplication, which are typically used for optimization or eigen value decomposition. Again, we use SparseDataset for mutable operations before we convert it to SparseMatrix.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: This isn't verbose. It is self-documenting.

math/src/main/java/smile/math/matrix/SparseMatrix.java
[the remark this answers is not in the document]
tdunning: Will do.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: Will switch to accept.

math/src/main/java/smile/math/matrix/SparseMatrix.java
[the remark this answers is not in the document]
tdunning: I removed the access to private data structures and added comments on benchmarks to the class javadoc.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: Done.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: A method like foreachNonZero needs to be public because it will be used by applications. That is the way that the code currently is.

But IntelliJ is analyzing just the Smile code and it notices that the method is only used in the SparseMatrixTest class and further notices that this test case is in the same package as the original code. It therefore incorrectly flags the definition of the function as having broader access rights than are necessary. Thus, the SuppressWarnings.

I could suppress the warning across the entire project for my own use, but that would leave the warnings visible to others who work on the code using similar toolsets. I think it is better to leave the warning levels at the default level so I see what everybody else will see. At that point it is good practice to mark some warnings as spurious so that any real warnings will be visible.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: In application code, the javadoc won't carry across, but the name will. Having the Nonzero as part of the name is important for readers of the code to understand the function.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: Where is this documented?

tdunning: Other kinds of matrices are mutable. Why only make SparseMatrix immutable?

The general feeling across the implementations I have seen (in R, in Arrow, in Pandas) and the implementations I have worked on in the past (many) is that mutation of a SparseMatrix is only a problem for brittle data structures and only then if the pattern of sparsity is changed.

Leaving the set method on the element makes this safe.

It is also very useful for operations such as in-place squaring or taking the absolute value.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: Done.

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: Done

math/src/main/java/smile/math/matrix/SparseMatrix.java
tdunning: You have just made pronouncements so far. Without technical justification, those are just opinions.

What is the technical rationale? I have provided context from the many other sparse matrix packages I have worked on that mutation on sparse matrices is only a problem due to brittleness of typical csr or rsr representations. This idiom prevents that problem.

haifengl:
Have you update the PR with new commits?

tdunning:
Not yet. Was waiting for a bit more convergence.

I will take your comment as that convergence.

tdunning:
Pushed update. Not all reviews are resolved yet, but a majority have been.

haifengl:
Thanks!

tdunning:
Any time.

twotabbies:
Removing SparseMatrix.values() seems to have broken plot/src/main/java/smile/plot/SparseMatrixPlot.java
[error] /usr/local/google/home/wjr/smile-twotabbies/smile/plot/src/main/java/smile/plot/SparseMatrixPlot.java:117:1: cannot find symbol
[error]   symbol:   method values()
[error]   location: variable sparse of type smile.math.matrix.SparseMatrix
[error]         for (double z : sparse.values()) {
[error] /usr/local/google/home/wjr/smile-twotabbies/smile/plot/src/main/java/smile/plot/SparseMatrixPlot.java:130:1: cannot find symbol
[error]   symbol:   method values()
[error]   location: variable sparse of type smile.math.matrix.SparseMatrix
[error]         for (double z : sparse.values()) {

tdunning:
Would you like for me to rewrite those parts of SparseMatrixPlot?

Or restore the values() method?

The uses of values() seem quite limited ... mostly for range testing.  Such
uses are easily replaced.

tdunning:
See
https://github.com/tdunning/smile/commit/b37629118b744eca7407f79da9837e3b89a107ba

for a proposed set of changes.

Issue #257, opened 2017-12-11
Training a Random Forest with a big dataset seems very slow


gcoter:
## Expected behavior
I have a dataset containing 21 000 000 instances. Each instance has 37 attributes (20 NominalAttribute, 17 NumericAttribute) and one binary label. My goal is to train a Random Forest with this dataset. As it is supposed to be used in a production setting, I need the training to take less than a few hours.

## Actual behaviour
I have benchmarked Smile’s implementation with Scikit Learn’s. Here are my current results:

- Scikit Learn: the training takes about 14 minutes.
- Smile: I provide below my logs.

```
[2017-12-08 19:31:27,927] [INFO] RfTrainer:78 main - Training...
[2017-12-08 19:31:27,928] [INFO] RfTrainer:79 main - Train X: (20932649,37); Train Y: (20932649,)
[2017-12-09 21:47:00,314] [INFO] RandomForest$TrainingTask:406 Thread-8 - Random forest tree OOB size: 20928906, accuracy: 99.90%
[2017-12-09 22:08:38,313] [INFO] RandomForest$TrainingTask:406 Thread-40 - Random forest tree OOB size: 20932480, accuracy: 99.90%
[2017-12-09 22:11:59,334] [INFO] RandomForest$TrainingTask:406 Thread-15 - Random forest tree OOB size: 20929089, accuracy: 99.90%
[2017-12-09 22:12:45,533] [INFO] RandomForest$TrainingTask:406 Thread-14 - Random forest tree OOB size: 20932361, accuracy: 99.90%
[2017-12-09 22:17:55,780] [INFO] RandomForest$TrainingTask:406 Thread-23 - Random forest tree OOB size: 20931951, accuracy: 99.90%
[2017-12-09 22:26:24,395] [INFO] RandomForest$TrainingTask:406 Thread-17 - Random forest tree OOB size: 20932623, accuracy: 99.90%
[2017-12-09 22:26:24,401] [ERROR] RandomForest:598 main - Failed to train random forest on multi-core
java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError: Java heap space
	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.util.concurrent.FutureTask.get(FutureTask.java:192)
	at smile.util.MulticoreExecutor.run(MulticoreExecutor.java:107)
	at smile.classification.RandomForest.<init>(RandomForest.java:596)
	at smile.classification.RandomForest.<init>(RandomForest.java:504)
[...]

Issue #70, opened 2016-03-02
DecisionTree and RegressionTree uses too much memory space


myui:
I found that `DecisionTree` and `RegressionTree` implementation use too large heap space for constructing trees as `split` method call depth grows because `trueSamples` and `falseSamples` is not released for GC until each `split` method call returns.

 https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/classification/DecisionTree.java#L655
https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/regression/RegressionTree.java#L599

Other RandomForest implementation uses `bag[N]`, in which bag[i] represents `sample[bag[i]]++`, instead of `samples[N]` and `bag.length` is reduced as the tree depth grows. In smile, `N` is the number of training examples and it consumes `O(2N * depth)` in split.

Releasing `samples` and `trueSamples` as soon as possible helps a little.
https://github.com/myui/hivemall/pull/259/files

Here is a condition that OOM happened due to recursive node splits.

```
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:705)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:701)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:701)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:714)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:714)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:701)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:714)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:714)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:714)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:701)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:701)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:701)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:714)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:701)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:701)
    at smile.regression.RegressionTree$TrainNode.split(RegressionTree.java:701)
[...]

PR #448, opened 2019-08-06
Speed up DecisionTree training by partitioning the order arrays

twotabbies:
This should help somewhat with https://github.com/haifengl/smile/issues/257 and https://github.com/haifengl/smile/issues/70

### DIFF
--- a/core/src/main/java/smile/classification/DecisionTree.java
+++ b/core/src/main/java/smile/classification/DecisionTree.java
@@ -18,6 +18,7 @@ package smile.classification;
 import java.io.Serializable;
 import java.util.*;
 import java.util.concurrent.Callable;
+import java.util.function.IntPredicate;
 import smile.data.Attribute;
 import smile.data.NominalAttribute;
 import smile.data.NumericAttribute;
@@ -135,10 +136,19 @@ public class DecisionTree implements SoftClassifier<double[]> {
      */
     private int mtry;
     /**
-     * The index of training values in ascending order. Note that only numeric
-     * attributes will be sorted.
+     * An index of training values. Initially, order[i] is a set of indices that iterate through the
+     * training values for attribute i in ascending order. During training, the array is rearranged
+     * so that all values for each leaf node occupy a contiguous range, but within that range they
+     * maintain the original ordering. Note that only numeric attributes will be sorted; non-numeric
+     * attributes will have a null in the corresponding place in the array.
      */
     private transient int[][] order;
+
+    /**
+     * An index of training values that maps their current position in the {@link #order} arrays to
+     * their original locations.
+     */
+    private transient int[] originalOrder;
 
     /**
      * Trainer for decision tree classifiers.
@@ -389,15 +399,25 @@ public class DecisionTree implements SoftClassifier<double[]> {
          * sampling with replacement.
          */
         int[] samples;
+        /**
+         * The lower bound (inclusive) in the order array of the samples belonging to this node.
+         */
+        int low;
+        /**
+         * The upper bound (exclusive) in the order array of the samples belonging to this node.
+         */
+        int high;
 
         /**
          * Constructor.
          */
-        public TrainNode(Node node, double[][] x, int[] y, int[] samples) {
+        public TrainNode(Node node, double[][] x, int[] y, int[] samples, int low, int high) {
             this.node = node;
             this.x = x;
             this.y = y;
             this.samples = samples;
+            this.low = low;
+            this.high = high;
         }
 
         @Override
@@ -415,6 +435,14 @@ public class DecisionTree implements SoftClassifier<double[]> {
              */
             int n;
             /**
+             * The lower bound (inclusive) in the order array of the samples belonging to this node.
+             */
+            int low;
+            /**
+             * The upper bound (exclusive) in the order array of the samples belonging to this node.
+             */
+            int high;
+            /**
              * The sample count in each class.
              */
             int[] count;
@@ -423,12 +451,14 @@ public class DecisionTree implements SoftClassifier<double[]> {
              */
             double impurity;
             /**
-             * The index of variables for this task.
+             * The variable for this task to attempt to split on.
              */
             int j;
 
-            SplitTask(int n, int[] count, double impurity, int j) {
+            SplitTask(int n, int low, int high, int[] count, double impurity, int j) {
                 this.n = n;
+                this.low = low;
+                this.high = high;
                 this.count = count;
                 this.impurity = impurity;
                 this.j = j;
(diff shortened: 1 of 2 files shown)


haifengl:
Thanks!

core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -744,13 +759,13 @@ private double impurity(int[] count, int n) {
 
         switch (rule) {
             case GINI:
-                impurity = 1.0;
+                double squared_sum = 0;
```
myui: @haifengl Did you review this modification?

core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -982,6 +997,119 @@ public DecisionTree(Attribute[] attributes, double[][] x, int[] y, int maxNodes,
...
+
+    /**
+     * Modifies an array in-place by partitioning the range from low (inclusive) to high
+     * (exclusive) so that all elements i for which left[i - low] is true come before all elements
+     * for which it is false, but element ordering is otherwise preserved. The number of true
+     * elements in left must equal split-low. buffer is scratch space large enough to hold all
+     * elements for which left is false.
+     */
+    private void partitionArray(int[] a, int low, int split, int high, IntPredicate goesLeft, int[] buffer) {
+        int n = high - low;
```
myui: `n` is not used.

core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -648,58 +671,47 @@ public boolean split(PriorityQueue<TrainNode> nextSplits) {
                 throw new IllegalStateException("Split a node with invalid feature.");
             }
 
-            int n = x.length;
+            int n = high - low;
```
myui: `n` is not used.

core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -982,6 +997,119 @@ public DecisionTree(Attribute[] attributes, double[][] x, int[] y, int maxNodes,
...
+        int k = 0;
+        for (int i = low; i < high; i++) {
+            if (goesLeft.test(a[i])) {
+                a[j++] = a[i];
+            } else {
+                buffer[k++] = a[i];
+            }
+        }
+        if (k != high - split || j != split) {
+            System.err.println("Messed up partition: " + low + ".." + split + ".." + high + " ended up splitting at " + j);
```
myui: Better to throw IllegalStateException for fail-fast if this condition should not be met.

myui:
@haifengl @twotabbies 

I'm not yet fully understand this approach but is [partitionOrder](https://github.com/haifengl/smile/pull/448/files#diff-aa1c5dce2c959f46efdcb90067706154R725) considers recursion in [split](https://github.com/haifengl/smile/pull/448/files#diff-aa1c5dce2c959f46efdcb90067706154R732)?

In [trueChild.split(null)](https://github.com/haifengl/smile/pull/448/files#diff-aa1c5dce2c959f46efdcb90067706154R732), the destructive `partitionOrder` method modified `order` field. Then, [falseChild.findBestSplit()](https://github.com/haifengl/smile/pull/448/files#diff-aa1c5dce2c959f46efdcb90067706154R737) uses modified `order` using `split/high`.

`split(null)` (depth-first split) is not used now as discussed in https://github.com/haifengl/smile/pull/359 though.

core/src/main/java/smile/classification/DecisionTree.java
twotabbies: This is supposed to be equivalent - both old and new compute $1 - \sum_i c_i^2 / n^2$, but the old approach has a division per class, while the new has only a single division.

core/src/main/java/smile/classification/DecisionTree.java
twotabbies: Thanks!

core/src/main/java/smile/classification/DecisionTree.java
twotabbies: Thanks!

core/src/main/java/smile/classification/DecisionTree.java
twotabbies: Sigh. There's always one left-over bit of debugging that should have been removed. The unit tests do pass.

twotabbies:
> @haifengl @twotabbies
> 
> I'm not yet fully understand this approach but is [partitionOrder](https://github.com/haifengl/smile/pull/448/files#diff-aa1c5dce2c959f46efdcb90067706154R725) considers recursion in [split](https://github.com/haifengl/smile/pull/448/files#diff-aa1c5dce2c959f46efdcb90067706154R732)?
> 
> In [trueChild.split(null)](https://github.com/haifengl/smile/pull/448/files#diff-aa1c5dce2c959f46efdcb90067706154R732), the destructive `partitionOrder` method modifies `order` field. Then, [falseChild.findBestSplit()](https://github.com/haifengl/smile/pull/448/files#diff-aa1c5dce2c959f46efdcb90067706154R737) uses modified `order` using `split/high`.
> 
> `split(null)` (depth-first split) is not used now as discussed in #359 though.

This is fine. partitionOrder() has partitioned the order arrays so that the true child owns low...split-1, and the false child owns split...high-1. A node and (importantly) all its descendants can modify only the range that it owns, so whether the true node is descended immediately (depth-first) or later (when pulled from nextSplits) it won't affect the expansion of the false node.

twotabbies:
Thanks for the code comments. Since this has been merged already, I'll put together another pull request with the modifications.

myui:
hmm, `(6)` uses `order` modified at `(3)` but range of modification is safe for (6).
```
-- level1
(1) partitionOrder(low1, split1, high1, ..)
(2) trueChild1.findBestSplit() -> findBestSplit(.., low1, split1, ..)
  -- level2
  (3) partitionOrder(low1, split2, split1, ..)
  (4) trueChild2.findBestSplit() -> findBestSplit(.., low1, split2, ..)
  (5) falseChild1.findBestSplit() -> findBestSplit(.., split2, high1, ..)
(6) falseChild1.findBestSplit() -> findBestSplit(.., split1, high1, ..)
```

Suppose best-first split as follows 

```
      (1)
     /   \
   (2)   (2)'
  /  \   /  \
(3) (3)' (4) (4)
    /  \
  (5) (5)

1. root splits
2. left side of 2nd level splits
3. right side of 3rd level (3) splits
4. right side of 2nd level splits 
```

level3 and level2 uses different low/high range but never overlaps, and thus destructive modifications to `order/originalOrder` field are safe (?).

```
-- level3 2nd leftmost node 
(1) node3.partitionOrder()
(2) trueChild3.findBestSplit() 
(3) falseChild3.findBestSplit() 
-- level2 rightmost node
(4) node2.partitionOrder() -- uses order/originalOrder modified in (1)
(5) trueChild2.findBestSplit()
(6) falseChild2.findBestSplit()
```

core/src/main/java/smile/classification/DecisionTree.java
myui: hmm, right. I have little concern about overflow at `squared_sum += ` though.

PR #449, opened 2019-08-07
Prune redundant branches in DecisionTrees

myui:
I found that smile's decision tree produces a lot of redundant branches as seen in 
https://user-images.githubusercontent.com/1163783/27733149-8a1927d8-5dcf-11e7-873b-9f0d37e19733.png

So, this PR prune redundant branches as follows:

Before: 
```javascript
if( sibsp <= 2.5 ) {
  if( age <= 15.5 ) {
    if( age <= 0.5 ) {
      if( sibsp <= 1.5 ) {
        if( parch <= 1.5 ) {
          no;
        } else  {
          no;
        }
      } else  {
        yes;
      }
    } else  {
      if( age <= 11.5 ) {
        if( fare <= 14.33125 ) {
          yes;
        } else  {
          yes;
        }
      } else  {
        yes;
      }
    }
  } else  {
    no;
  }
} else  {
  no;
}
```

After:
```javascript
if( sibsp <= 2.5 ) {
  if( age <= 15.5 ) {
    if( age <= 0.5 ) {
      if( sibsp <= 1.5 ) {
        no;
      } else  {
        yes;
      }
    } else  {
      yes;
    }
  } else  {
    no;
  }
} else  {
  no;
}
```

This is a back-port of [my modification](https://github.com/apache/incubator-hivemall/pull/198) to smile.

### DIFF
--- a/core/src/main/java/smile/classification/DecisionTree.java
+++ b/core/src/main/java/smile/classification/DecisionTree.java
@@ -973,9 +973,6 @@ public class DecisionTree implements SoftClassifier<double[]> {
             }
         }
 
-        // Priority queue for best-first tree growing.
-        PriorityQueue<TrainNode> nextSplits = new PriorityQueue<>();
-
         int n = y.length;
         int[] count = new int[k];
         if (samples == null) {
@@ -1003,21 +1000,28 @@ public class DecisionTree implements SoftClassifier<double[]> {
         root = new Node(Math.whichMax(count), posteriori);
         
         TrainNode trainRoot = new TrainNode(root, x, y, samples, 0, originalOrder.length);
-        // Now add splits to the tree until max tree size is reached
-        if (trainRoot.findBestSplit()) {
-            nextSplits.add(trainRoot);
-        }
-
-        // Pop best leaf from priority queue, split it, and push
-        // children nodes into the queue if possible.
-        for (int leaves = 1; leaves < this.maxNodes; leaves++) {
-            // parent is the leaf to split
-            TrainNode node = nextSplits.poll();
-            if (node == null) {
-                break;
+        if(maxNodes == Integer.MAX_VALUE) {// depth-first split
+            if (trainRoot.findBestSplit()) {
+                trainRoot.split(null);
             }
+        } else {// best-first split
+            // Priority queue for best-first tree growing.
+            PriorityQueue<TrainNode> nextSplits = new PriorityQueue<>();
 
-            node.split(nextSplits); // Split the parent node into two children nodes
+            // Now add splits to the tree until max tree size is reached
+            if (trainRoot.findBestSplit()) {
+                nextSplits.add(trainRoot);
+            }
+            // Pop best leaf from priority queue, split it, and push
+            // children nodes into the queue if possible.
+            for (int leaves = 1; leaves < this.maxNodes; leaves++) {
+                // parent is the leaf to split
+                TrainNode node = nextSplits.poll();
+                if (node == null) {
+                    break;
+                }
+                node.split(nextSplits); // Split the parent node into two children nodes
+            }
         }
 
         this.order = null;
--- a/core/src/main/java/smile/classification/DecisionTree.java
+++ b/core/src/main/java/smile/classification/DecisionTree.java
@@ -324,6 +324,12 @@ public class DecisionTree implements SoftClassifier<double[]> {
             this.posteriori = posteriori;
         }
 
+        private void markAsLeaf() {
+            this.splitFeature = -1;
+            this.splitValue = Double.NaN;
+            this.splitScore = 0.0;
+        }
+
         /**
          * Evaluate the regression tree over an instance.
          */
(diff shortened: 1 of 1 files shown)


myui:
BTW, I'm not sure whether we should [avoid splitScore of pruned nodes](https://github.com/haifengl/smile/pull/449/files#diff-aa1c5dce2c959f46efdcb90067706154L730) or not.

Even for pruned nodes, splitScore may be useful for feature importance.

myui:
Ah, this bottom-up pruning scheme only work for depth-first tree splitting and now smile only supports a best-first split using priority queue.
https://github.com/haifengl/smile/pull/359#issuecomment-510336788

haifengl:
Thanks. Can you please resolve the conflicts?

myui:
Resolved conflicts.

haifengl:
If this works only with depth-first tree splitting, can we really merge this? Thanks

myui:
It also works for best-first splitting without any drawbacks. When supporting depth-first splitting, it would help. 

The best-fist and depth-first creates a same tree if `max number of leaf nodes` is not specified. And it's better to support depth-first splitting again. Pushed to support depth-first split as well. https://github.com/haifengl/smile/pull/449/commits/7f96bea90ab743a1013b81abfe01ee0618644068

core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -981,21 +1001,28 @@ public DecisionTree(Attribute[] attributes, double[][] x, int[] y, int maxNodes,
...
+            }
+            // Pop best leaf from priority queue, split it, and push
+            // children nodes into the queue if possible.
+            for (int leaves = 1; leaves < this.maxNodes; leaves++) {
+                // parent is the leaf to split
+                TrainNode node = nextSplits.poll();
+                if (node == null) {
+                    break;
+                }
+                node.split(nextSplits); // Split the parent node into two children nodes
```
twotabbies: I think this should be
if (!node.split(nextSplits)) { leaves--; }
because if the node was split into two leaves with the same label and the split was aborted because of that, then the split shouldn't count towards the tree's leaf limit.

myui: You're right. 

The same problem exists in the current master as seen in 
https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/classification/DecisionTree.java#L997

twotabbies: True - but if I've followed the logic correctly, I think that in the current master node.split() will never return false; after your change it will return false in some cases.

Why I think it won't return false at the moment:
TrainNode.split() can only be called if node.splitFeature >= 0. It will return false only if applying the split specified by (node.splitFeature, node.splitValue) results in either the true branch or the false branch having a size < nodeSize. If both the true and false branches are >= nodeSize, then it must (currently) return true. So in order for it to return false, it needs to be holding a node having a split criterion that produces a too-small branch on one side or the other. That split criterion needs to have been generated by one of the calls to findBestSplit(n, ...) - but https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/classification/DecisionTree.java#L595 and https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/classification/DecisionTree.java#L633 won't approve such a criterion; in order for splitFeature/splitValue to get set, the code has to get past one or the other of those checks, meaning that both the true and false branches will be large enough... so split() can't currently return false.

Aside: those two checks in turn seem to imply that the checks at https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/classification/DecisionTree.java#L507, https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/classification/DecisionTree.java#L727 and https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/classification/DecisionTree.java#L736 (which shortcut, or refuse to call, findBestSplit() if the candidate's node size <= nodeSize) could be tightened up to check "< 2 * nodeSize" - unless you have at least 2*nodeSize units to spread around, at least one branch is going to be <nodeSize. This would speed up tree construction slightly.

myui: I'm not 100% confident [this condition](https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/classification/DecisionTree.java#L711) never be happen while your investigation seems right.

Anyway, returning value of `split()` should be meaningful without assuming it would not return false. 

`leaves--` should be inserted when split() returns false. +1 on this option.
Or, better to throw an exception if unexpected behavior (i.e., return false) happens.

> check "< 2 * nodeSize" - unless you have at least 2*nodeSize units to spread around, at least one branch is going to be <nodeSize.

Yeah, it's better to be ensured.

I'm doing it in [my smile modification](https://github.com/apache/incubator-hivemall/pull/198/commits/294ce93fd4ac2c855da5af74fccfc6213456d5f8#diff-2c14597fd8cb87295b3206710c62eeb4R1148).

myui:
Fixed to do `if (!node.split(nextSplits)) { leaves--; }`
https://github.com/haifengl/smile/pull/449/commits/d3f56b82ab90726c95fad492c93febbfbd51a599

PR #453, opened 2019-08-15
Various SparseMatrix improvements

eddspencer:
This PR has been done to optimise the memory and runtime needs of using SparseMatrix objects. It contains the following changes:

### Floats for SparseMatrix objects instead of doubles
This saves half the memory when doing computations that involve SparseMatrix objects (with the drawback of losing some decimal places). For that purpose, a `float[] xF` attribute is added, and used if `boolean doubles` is set to true when instantiating a new SparseMatrix.
The `get()` method has been adapted to handle both cases, and a `set()` method have been added.

### SparseMatrix attributes made public
Using `get()` to query the SparseMatrix values doesn’t scale well. The `nrows`, `ncols`, `colIndex` and `rowIndex` attributes are hence made public for direct and quick access. 

### Optimising the aat() method
`aat()` returns A*A^T, provided a SparseMatrix object A. The result is a symmetric matrix. We can save about half the memory by not returning the whole matrix but only the lower triangular part. For that purpose, a `lowerTriangularSparseAAT()` method have been added which returns a SparseMatrix object containing the lower triangular elements of the AAT operation.

### Condensed proximity matrix constructor in Linkage
`Linkage` currently only accepts 2D dense arrays, and creates a 1D condensed array (with only the lower triangular elements) out of it. This is not optimal in terms of memory if the 1D condensed array can be provided directly.
A `init(float[] proximity)` method has been added so that the condensed proximity matrix can directly be provided to the HAC algorithm, without going through the 2D dense array step.
This modification has also been made in the `UPGMALinkage` constructor, and could be done for any other linkage method.

To account for issues when the condensed arrays becomes large, the `size` attribute of `Linkage` is changed to `long`, and casted to `int` when it’s used outside of the class. Without this change, computations such as `size * (size+1)/ 2` can fail because the operations temporarily exceeds the maximum integer size (although the final result may not).

A `getSize(int sizeCondensed)` method has been added that gives the number of elements in the 2D dense array of the corresponding condensed array.

### DIFF
--- a/core/src/main/java/smile/clustering/linkage/Linkage.java
+++ b/core/src/main/java/smile/clustering/linkage/Linkage.java
@@ -30,7 +30,7 @@ package smile.clustering.linkage;
  */
 public abstract class Linkage {
     /** The data size. */
-    int size;
+    long size;
 
     /**
      * Linearized proximity matrix to store the pair-wise distance measure
@@ -45,7 +45,7 @@ public abstract class Linkage {
     /** Initialize the linkage with the lower triangular proximity matrix. */
     void init(double[][] proximity) {
         size = proximity.length;
-        this.proximity = new float[size * (size+1) / 2];
+        this.proximity = new float[proximity.length * (proximity.length+1) / 2];
 
         // row wise
         /*
@@ -65,16 +65,31 @@ public abstract class Linkage {
         }
     }
 
+    void init(float[] proximity) {
+        size = getSize(proximity.length);
+        this.proximity = proximity;
+    }
+
+    private int getSize(int sizeCondensed) {
+        int i = 0;
+        int result = 0;
+        while (result != sizeCondensed) {
+            i++;
+            result += i;
+        }
+        return i;
+    }
+
     int index(int i, int j) {
         // row wise
         // return i > j ? i*(i+1)/2 + j : j*(j+1)/2 + i;
         // column wise
-        return i > j ? proximity.length - (size-j)*(size-j+1)/2 + i - j : proximity.length - (size-i)*(size-i+1)/2 + j - i;
+        return (int) (i > j ? proximity.length - (size-j)*(size-j+1)/2 + i - j : proximity.length - (size-i)*(size-i+1)/2 + j - i);
     }
 
     /** Returns the proximity matrix size. */
     public int size() {
-        return size;
+        return (int) size;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


haifengl:
SparseMatrix is designed for efficient matrix-vector and matrix-matrix multiplication. In general, we should not use `get` method directly unless testing/debugging. `get` is slow for any sparse matrix design. It is really a bad idea to make member variables like `colIndex` etc public. It breaks the encapsulation of the design.

haifengl:
Thanks for your contributions. But most of your changes are against our original design choices. It would be better for you to communicate first before making a PR in the future. Thanks again.

eddspencer:
Sorry I have not gotten back sooner, would you be open to discussing these changes in more detail? 

The main issue these changes are trying to address is one of memory usage, we have a square sparse matrix of order 20k which is then converted to a dense matrix to use the UPGMA linkage. We are using the member variables like `colIndex` to create the proximity matrix with only the lower triangular values populated (which is sufficient for our needs) to save memory.

The change to allow `float` in the SparseMatrix also has massive memory saving implications, and I do not know of a better way to to do this in Java without creating a different class. Scala specialised generics could be used here I think.

Would it help if some of these changes are removed and a PR is made with a sub-set, you mention `most of your changes are against our original design choices`?

eddspencer:
Many thanks for your quick reply! Would you have any alternative suggestions to using a SparseMatrix here, or are you suggesting just using java arrays for the triangular matrix?

Regarding the memory usage I agree that it is simple enough with today's hardware but we have (currently) 7 of these computations that need repeating every 20 minutes or so, using AWS for this means we are conscious of pricing and therefore looking to reduce the memory usage where possible. 

To that end what about the edits to the Linkage classes? It would be ideal to be able to generate the `float[] proximity` array ourselves and just pass that into the constructor rather than (momentarily) having a `double[][]` copy in memory as well.

haifengl:
In older versions of Smile, hierarchical clustering takes a double[][] as input. We switch to Matrix as it is faster indeed. For short term, you can fork the older version and change it to float (please no PR). For longer term, we need add TriangularMatrix and also support float version for all matrix implementation. It will be a lot of effort and takes some time.

haifengl:
We actually store data in `float` in `Linkage` and we store only half of the matrix to save space. The `Linkage` constructor takes a double[][] as input though. I can add another constructor taking `float[]` as input that already save data in column-wise lower-triangular format (no diagonals).

haifengl:
I add a new constructor to Linkage to take `float[]`. See javadoc for details.

eddspencer:
Thanks for adding that to the Linkage.

PR #454, opened 2019-08-21
Reduce size of DecisionTrees: remove unnecessary members; clear unuse…

twotabbies:
This uses three techniques to reduce the in-RAM (and on-disk, when serialised) size of DecisionTrees.
1. Remove two members that are used only during training.
2. Clear the "importance" array on interior nodes, as it's used only on leaves and can be quite large if there are many classes.
3. As suggested by myui, remove splits where both sides have the same label.

A couple of important notes:
- This conflicts with https://github.com/haifengl/smile/pull/449. It does a more general job of pruning by waiting until the tree is fully constructed so that multiple levels that all have the same label can be merged.
- This changes the list of non-transient members of Node, which will affect its autogenerated serialVersionUID, so any serialised DecisionTrees generated from old code will not be readable in new code and vice versa. The changes are such that I believe that inserting an explicit serialVersionUID equal to the current autogenerated one will actually give us forward and backward compatibility, but whether to do that is a decision that I don't want to make.

On the "airline-100k" benchmark, this reduces the serialised size of the RandomForest from ~6M to ~2.4M, mostly due to the pruning; on an internal benchmark, it goes from ~50M to ~26M, mostly due to the removal of importance arrays on interior nodes.

### DIFF
--- a/core/src/main/java/smile/classification/DecisionTree.java
+++ b/core/src/main/java/smile/classification/DecisionTree.java
@@ -316,6 +316,14 @@ public class DecisionTree implements SoftClassifier<double[]> {
             this.posteriori = posteriori;
         }
 
+        private void markAsLeaf() {
+            this.splitFeature = -1;
+            this.splitValue = Double.NaN;
+            this.splitScore = 0.0;
+            this.trueChild = null;
+            this.falseChild = null;
+        }
+
         /**
          * Evaluate the regression tree over an instance.
          */
@@ -691,9 +699,7 @@ public class DecisionTree implements SoftClassifier<double[]> {
             }
 
             if (tc < nodeSize || fc < nodeSize) {
-                node.splitFeature = -1;
-                node.splitValue = Double.NaN;
-                node.splitScore = 0.0;
+                node.markAsLeaf();
                 return false;
             }
 
@@ -711,13 +717,18 @@ public class DecisionTree implements SoftClassifier<double[]> {
             int[] buffer = new int[high - split];
             partitionOrder(low, split, high, goesLeft, buffer);
 
+            int leaves = 0;
             TrainNode trueChild = new TrainNode(node.trueChild, x, y, samples, low, split);
             if (tc > nodeSize && trueChild.findBestSplit()) {
                 if (nextSplits != null) {
                     nextSplits.add(trueChild);
                 } else {
-                    trueChild.split(null);
+                    if (trueChild.split(null) == false) {
+                        leaves++;
+                    }
                 }
+            } else {
+                leaves++;
             }
 
             TrainNode falseChild = new TrainNode(node.falseChild, x, y, samples, split, high);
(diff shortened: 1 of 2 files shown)


core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -997,6 +985,8 @@ public DecisionTree(Attribute[] attributes, double[][] x, int[] y, int maxNodes,
             node.split(nextSplits); // Split the parent node into two children nodes
         }
 
+        pruneRedundantLeaves(root);
```
myui: I think post pruning should be used for best-first splitting at https://github.com/haifengl/smile/pull/449/files#diff-aa1c5dce2c959f46efdcb90067706154R1025 when https://github.com/haifengl/smile/pull/454 is merged.

https://github.com/haifengl/smile/pull/454 does pruning while building a tree in depth-first and thus the maximum memory consumption is reduced.

And, post-pruning could be generalized to do the simple pruning or to do reduced error (say more aggressive) pruning https://github.com/haifengl/smile/pull/360 by a configuration.

twotabbies: You're right; we can do pruning while building the tree, and then also do a post-pass to re-merge any redundant splits, which can still happen in best-first and depth-first splitting.

Reduced error pruning is worth exploring later.

haifengl:
@twotabbies and @myui is it possible that we merge you two's PRs into one? Thanks.

myui:
@haifengl If you merge one to master, we can rebase to master.

core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -980,23 +988,34 @@ public DecisionTree(Attribute[] attributes, double[][] x, int[] y, int maxNodes,
...
+            for (int leaves = 1; leaves < this.maxNodes; leaves++) {
+                // parent is the leaf to split
+                TrainNode node = nextSplits.poll();
+                if (node == null) {
+                    break;
+                }
+                if (!node.split(nextSplits)) { // Split the parent node into two children nodes
+                    leaves--;
+                }
+            }
```
myui: The current `pruneRedundantLeaves(root);` can be moved to this line.

myui: I think cleaning out of `posteriori` could be done in https://github.com/apache/incubator-hivemall/blob/master/core/src/main/java/hivemall/smile/classification/DecisionTree.java#L861

It's better because it's a way efficient for depth-first splitting and also works for best-first splitting.

myui: You're right. With post-pruning and best-first split, it should be

```
if(nextSplits != null) {
  node.posteriori = null;
}
```

Then, post-pruning is not required for depth-first splitting. And, it would work fine for best-first splitting.

twotabbies: I've made this change.

myui: Thanks. confirmed.

myui:
LGTM.

haifengl:
Thanks!

PR #461, opened 2019-08-30
Avoid constant features in findBestSplit

myui:
This PR avoids constant features in finding best split.

Here is [a quote from scikit](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/tree/_splitter.pyx#L1455).

```
Skip the CPU intensive evaluation of the impurity criterion for features that were already detected as constant (hence not suitable for good splitting) by ancestor nodes and save the information on newly discovered constant features to spare computation on descendant nodes.
```

Not ideal but works well enough for multi-threaded execution as well.

This is a backport from [my smile modification](https://github.com/apache/incubator-hivemall/pull/198).

### DIFF
--- a/core/src/main/java/smile/classification/DecisionTree.java
+++ b/core/src/main/java/smile/classification/DecisionTree.java
@@ -559,7 +559,7 @@ public class DecisionTree implements SoftClassifier<double[]> {
             } else {
 
                 final int[] constFeatures_;
-                synchronized (constFeatures) {
+                synchronized (this.constFeatures) {// enforce to read up-to-date constFeature
                     constFeatures_ = this.constFeatures; // this.constFeatures may be replaced in findBestSplit but it's accepted
                 }
                 List<SplitTask> tasks = new ArrayList<>(mtry);
@@ -719,9 +719,9 @@ public class DecisionTree implements SoftClassifier<double[]> {
             }
 
             if(replaceCount + (countNaN > 0 ? 1: 0) <= 1) {
-                synchronized (constFeatures) {
-                    this.constFeatures = SmileUtils.sortedArraySet(constFeatures, j);
-                }
+                synchronized (this.constFeatures) { // enforce to read up-to-date constFeatures
+                    this.constFeatures = SmileUtils.sortedArraySet(this.constFeatures, j);
+                } // this.constFeatures in CPU cache is flushed
             }
 
             return splitNode;
--- a/core/src/main/java/smile/classification/DecisionTree.java
+++ b/core/src/main/java/smile/classification/DecisionTree.java
@@ -25,6 +25,7 @@ import smile.data.NumericAttribute;
 import smile.math.Math;
 import smile.sort.QuickSort;
 import smile.util.MulticoreExecutor;
+import smile.util.SmileUtils;
 
 /**
  * Decision tree for classification. A decision tree can be learned by
@@ -415,17 +416,29 @@ public class DecisionTree implements SoftClassifier<double[]> {
          * The upper bound (exclusive) in the order array of the samples belonging to this node.
          */
         int high;
+        /**
+         * Constant feature indices
+         */
+        int[] constFeatures;
 
         /**
          * Constructor.
          */
         public TrainNode(Node node, double[][] x, int[] y, int[] samples, int low, int high) {
+            this(node, x, y, samples, low, high, new int[0]);
+        }
+
+        /**
+         * Constructor.
+         */
+        public TrainNode(Node node, double[][] x, int[] y, int[] samples, int low, int high, int[] constFeatures) {
             this.node = node;
             this.x = x;
             this.y = y;
             this.samples = samples;
             this.low = low;
             this.high = high;
+            this.constFeatures = constFeatures;
         }
 
         @Override
@@ -527,8 +540,13 @@ public class DecisionTree implements SoftClassifier<double[]> {
                 Math.permutate(variables);
                 int[] falseCount = new int[k];
 
+                final int[] constFeatures_  = this.constFeatures; // this.constFeatures may be replaced in findBestSplit but it's accepted
+
                 // Random forest already runs on parallel.
                 for (int j = 0; j < mtry; j++) {
+                    if(Arrays.binarySearch(constFeatures_, j) >= 0) {
+                        continue; // skip constant features
+                    }
                     Node split = findBestSplit(n, low, high, count, falseCount, impurity, variables[j]);
                     if (split.splitScore > node.splitScore) {
                         node.splitFeature = split.splitFeature;
@@ -540,8 +558,15 @@ public class DecisionTree implements SoftClassifier<double[]> {
                 }
             } else {
 
+                final int[] constFeatures_;
+                synchronized (constFeatures) {
+                    constFeatures_ = this.constFeatures; // this.constFeatures may be replaced in findBestSplit but it's accepted
+                }
                 List<SplitTask> tasks = new ArrayList<>(mtry);
                 for (int j = 0; j < p; j++) {
+                    if(Arrays.binarySearch(constFeatures_, j) >= 0) {
+                        continue; // skip constant features
+                    }
                     tasks.add(new SplitTask(n, low, high, count, impurity, j));
                 }
 
@@ -558,6 +583,9 @@ public class DecisionTree implements SoftClassifier<double[]> {
                 } catch (Exception ex) {
                     int[] falseCount = new int[k];
                     for (int j = 0; j < p; j++) {
+                        if(Arrays.binarySearch(constFeatures_, j) >= 0) {
+                            continue; // skip constant features
+                        }
                         Node split = findBestSplit(n, low, high, count, falseCount, impurity, j);
                         if (split.splitScore > node.splitScore) {
                             node.splitFeature = split.splitFeature;
@@ -585,6 +613,8 @@ public class DecisionTree implements SoftClassifier<double[]> {
          */
         public Node findBestSplit(int n, int low, int high, int[] count, int[] falseCount, double impurity, int j) {
             Node splitNode = new Node();
+            double lastx = Double.NaN;
+            int replaceCount = 0, countNaN = 0;
 
             if (attributes[j].getType() == Attribute.Type.NOMINAL) {
                 int m = ((NominalAttribute) attributes[j]).size();
@@ -592,9 +622,17 @@ public class DecisionTree implements SoftClassifier<double[]> {
 
                 for (int i = low; i < high; i++) {
                     int o = originalOrder[i];
-                    trueCount[(int) x[o][j]][y[o]] += samples[o];
+                    double x_oj = x[o][j];
+                    if (Double.isNaN(x_oj)) {
+                        countNaN++;
+                        continue; // NaN should be avoided for split value. see (int)x_o
+                    }
+                    if (lastx != x_oj) {
+                        lastx = x_oj;
+                        replaceCount++;
+                    }
+                    trueCount[(int)x_oj][y[o]] += samples[o];
                 }
-
                 for (int l = 0; l < m; l++) {
                     int tc = Math.sum(trueCount[l]);
                     int fc = n - tc;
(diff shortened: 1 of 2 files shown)


core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -540,8 +558,15 @@ public boolean findBestSplit() {
                 }
             } else {
 
+                final int[] constFeatures_;
+                synchronized (constFeatures) {
```
myui: `synchronized` is used to sync CPU caches and for ordering.

If even other thread replaces `this.constFeatures`, it might not be seen in the current thread.
Missing some `j` in constFeatures by contention is allowed because it's an optional optimization.

twotabbies: This "synchronized" isn't necessary: at this point, no other thread can be accessing this.constFeatures. In fact, you don't need constFeatures_ at all, as there are no other threads accessing this.constFeatures during the time when you're using constFeatures_.

Once MulticoreExecutor.run() has completed, any modifications to this.constFeatures will be visible to the current thread: MulticoreExecutor invokes Future.get() which involves a lock acquisition and release, which means that all modifications that happen-before the acquisition are visible to this thread. The thread that did the work has to have placed the result in the Future, which means that everything it did happens-before the Future.get(), so if it changed this.constFeatures, it will be visible to this thread.

core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -654,20 +704,26 @@ public Node findBestSplit(int n, int low, int high, int[] count, int[] falseCoun
...
+                    prevy = y_o;
+                    trueCount[y_o] += samples[o];
                 }
             } else {
                 throw new IllegalStateException("Unsupported attribute type: " + attributes[j].getType());
             }
 
+            if(replaceCount + (countNaN > 0 ? 1: 0) <= 1) {
+                synchronized (constFeatures) {
+                    this.constFeatures = SmileUtils.sortedArraySet(constFeatures, j);
```
twotabbies: I found this line confusing, since it's referring to the same thing (this.constFeatures) by two different names (this.constFeatures, constFeatures) - leading me to wonder if they were two different entities at this point.

core/src/main/java/smile/classification/DecisionTree.java
```diff
@@ -654,20 +704,26 @@ public Node findBestSplit(int n, int low, int high, int[] count, int[] falseCoun
...
+                    prevx = x_oj;
+                    prevy = y_o;
+                    trueCount[y_o] += samples[o];
                 }
             } else {
                 throw new IllegalStateException("Unsupported attribute type: " + attributes[j].getType());
             }
 
+            if(replaceCount + (countNaN > 0 ? 1: 0) <= 1) {
+                synchronized (constFeatures) {
```
twotabbies: This synchronized block won't properly protect the modifications to constFeatures: two threads could replace it with different versions. To set this up, I think you need three threads...
Say thread A is running with j=1, B is running with j=2, and C is running with j=3. Initially, this.constFeatures = (say) CF1.
A sees that feature 1 is constant, and does
synchronized(CF1) { this.constFeatures = CF2; } where CF2 = CF1 + "1"
Some time later, B sees that feature 2 is constant - and after A did the above, it does (this is one possible legal execution)
syncronized(CF2) { this.constFeatures = CF3; } where CF3 = CF2 + "2".
Then thread C sees that feature 3 is constant, and does (this is also a legal execution)
synchronized(CF1) { this.constFeatures = CF4; } where CF4 = CF2 + "3". "2" has been lost.
Why did C do synchronized(CF1), not synchronized(CF2) or synchronized(CF3), and why did C base CF4 off CF2? Basically, until they acquire a lock, B and C are under no obligation to witness any modifications A has made - they might or they might not. So B happens to witness them, so it does synchronized(CF2) and modifies this.constFeatures to point to CF3. But C doesn't witness A's changes, or B's changes, and does syncronized(CF1). After it acquires the CF1 lock, it is then required to see modifications that were made while holding the lock - which were thread A's modifications of this.constFeatures to point to CF2. C isn't required to see B's modifications because those were made while holding a different lock (CF2)... so CF4 is based off CF2, and B's modification is lost.

I think that "synchronized(this)" is more reliable easier to understand, since the lock being used doesn't shift during execution.

core/src/main/java/smile/classification/DecisionTree.java
myui: I intended to use `synchronized` for CPU cache store buffer to flush and sync `constFeatures_` among threads. 

No need to be atomic paying synchronization costs but better to sync often.

core/src/main/java/smile/classification/DecisionTree.java
myui: > This synchronized block won't properly protect the modifications to constFeatures
It's intended.

No need to be atomic but synchronize `this.constFeature` in CPU caches when reading/writing `this.constFeature`.

core/src/main/java/smile/classification/DecisionTree.java
myui: > Once MulticoreExecutor.run() has completed, any modifications to this.constFeatures will be visible to the current thread: MulticoreExecutor invokes Future.get() which involves a lock acquisition and release, which means that all modifications that happen-before the acquisition are visible to this thread. 

True. It's just a timing issue of synchronization in the middle of findBestSplit or in the end of findBestSplit.

myui:
Thank you for your review. I'll revise them in my spare time.

core/src/main/java/smile/classification/DecisionTree.java
twotabbies: The real issue is that you're synchonizing on the object currently stored in this.constFeatures - which changes over time. To get consistent behaviour, you should synchronize on something that doesn't change over time; "this" is handy :-)

myui: synchronizing on `this.constFeatures` is enough since I'm using `synchronized(this.constFeatures)` for memory fence. Synchronizing changes by a particular thread to another threads. Using stale constFeatures does not have side effects.

core/src/main/java/smile/classification/DecisionTree.java
myui: Fixed in https://github.com/haifengl/smile/pull/461/commits/74ce0f768deba72f72ab55488911f9c365b458b8

core/src/main/java/smile/classification/DecisionTree.java
twotabbies: Synchronozing on this.constFeatures
- is harder to understand than just synchronizing on "this" because this.constFeatures' value can change, so the reader has to do a lot of logic to figure out the possible effects
- does not cover all cases (see my example above where a const feature is forgotten about)
Synchronization isn't a global memory barrier, and doesn't necessarily affect CPU caches at all (if a CPU core knows that the cache line holding the lock is not present in any other core's cache, say via https://en.wikipedia.org/wiki/MOESI_protocol, then the lock acquisition is done purely locally to that one core's cache; there's no flushing of anything).

Synchronizing on "this" makes the code more understandable, more correct, and more performant (since constant features will never be forgotten about).

myui: Again, no need to synchronized on `this`. No synchronization is required.

Java Memory Model enforces memory fence. See 
http://gee.cs.oswego.edu/dl/jmm/cookbook.html
```
EnterLoad is needed on entry to any synchronized block/method that performs a load. It is the same as LoadLoad unless an atomic instruction is used in MonitorEnter and itself provides a barrier with at least the properties of LoadLoad, in which case it is a no-op.
StoreExit is needed on exit of any synchronized block/method that performs a store. It is the same as StoreStore unless an atomic instruction is used in MonitorExit and itself provides a barrier with at least the properties of StoreStore, in which case it is a no-op.
ExitEnter is the same as StoreLoad unless atomic instructions are used in MonitorExit and/or MonitorEnter and at least one of these provide a barrier with at least the properties of StoreLoad, in which case it is a no-op.
```

myui: > this.constFeatures' value can change

It's accepted.

> Synchronization isn't a global memory barrier

No, Java `synchronized` enforces memory barrier.

haifengl:
I have implemented this idea in v2. Please check smile.base.cart. Thanks!

myui:
@haifengl which part? cannot see in https://github.com/haifengl/smile/blob/v2/core/src/main/java/smile/base/cart/CART.java

haifengl:
See Split.java line 64, CART.java line 281, 292, and 226 & 227. That it is.

myui:
It seems `pure` does not represents a constant or not  https://github.com/haifengl/smile/blob/v2/core/src/main/java/smile/base/cart/CART.java#L292

`j` is a split candidate or not.

Javadoc `pure true if all the samples in the split have the same value in the column.` is misleading.
https://github.com/haifengl/smile/blob/v2/core/src/main/java/smile/base/cart/CART.java#L267

`pure[j] is true if column j considered not to be a split candidate.` ?

myui:
It's a different concept to constant detection but confirmed that a similar idea (I'm not sure it works properly or not) is implemented in v2.

haifengl:
`imsplitable` is probably a better name. I choose `pure` as  it is short. `pure[j]` in a parent node is true in 3 cases:

- The column j in the node is constant
- The column j is not constant but after the split, the size of one or two children is less than nodeSize. We may call this "almost constant"
- The column j is not (almost) constant, but the impurity of node (GINI, RSS, etc.) doesn't decrease after the split.

No matter which case, the column j in the children nodes (split on another column) are still insplitable.

I think that this design is more general than just constant detection. And it is very efficient (no extra data scan at all).

myui:
> The column j is not (almost) constant, but the impurity of node (GINI, RSS, etc.) doesn't decrease after the split.

I'm concerned about this case.  

Does the column `j` never become split feature once findBestSplit returned an empty result for `j`?

Never goes the following lines for `unsplittable j` ?
https://github.com/haifengl/smile/blob/v2/core/src/main/java/smile/classification/DecisionTree.java#L177
https://github.com/haifengl/smile/blob/v2/core/src/main/java/smile/classification/DecisionTree.java#L220 

Purity for `j` can change as split goes on and `splitScore` may decrease.

myui:
Suppose that the column `j=1` returned an empty result (thus unsplittable[1]=true) and split happened for `j=2`. 

After that one of the child node can be split on `j=1` because purity changes by `j=2` splitting.

haifengl:
It is true that the node may become splittable on the column. But it is hard to argue that it is good thing. Note that the impurity is on y. If the node cannot reduce the impurity of y by splitting column j, it means that the correlation between column j and y is very low. Should we really care this noisy column? Would we just fit the coincidence on column j after we split on another column?

I did some experiments by turning on/off this design.

- For gradient tree boosting, there is no difference with cross validation or on test data. This is because gradient tree boosting uses very shallow trees.

- For AdaBoost and RandomForest, there are small difference as they use deep trees. However, the results are mixed. Sometimes, the error rate is a little bit lower. Sometimes, a little bit higher. It may be due to aforementioned lower correlation between unsplittable features and y.

myui:
> But it is hard to argue that it is good thing. 

True. At the same time, it's hard to argue that it's good thing or bad thing.

Correlation between column j and y and prediction accuracy are different types of argument. Even if correlation between column j and y is low, prediction accuracy using (j and) y can be improved for a certain case.

So, optimization using constants is more safe approach as same accuracy can be expected with or without optimization.

haifengl:
When I said "correlation", I meant it philosophically. Mutual information is a better term to describe what I try to say (not exactly same though).

It is very common that we want to leverage every bit of information even though we actually introduce more noise.

I will do more experiments to decide which approach is better empirically (if there are any statistical significance).

myui:
@haifengl 

`filter(i -> !unsplittable[i])` is done before column sampling. Then, size of [stream](https://github.com/haifengl/smile/blob/v2/core/src/main/java/smile/base/cart/CART.java#L326) remain `mtry`. 

I conservatively think that this unsplittable filtering should better be revised because elapsed time increase with this scheme. And, randomness of column sampling could get worse as `unsplittable` grows.

https://github.com/haifengl/smile/blob/v2/core/src/main/java/smile/base/cart/CART.java#L319

It may have good effects in accuracy in a certain case though.

haifengl:
It is intentional. So far, I don't notice any issues. As stated before,

- This design has no effects to gradient tree boost and adaboost as they use very shallow trees.
- It may be in effect for random forest as it uses deep trees. However, the experiments on many data shows that the unsplittable features are infrequent due to many reasons, e.g. best first splitting, good mutual information between predicators and response in general, the control of maxDepth, etc.
- In random forest for classification, mtry is often (very) small. I use unsplittable to focus on more informative features. It is about accuracy rather than speed. We are already very fast any way.

haifengl:
Close this for now. If we have issues, we can create a new topic.

Issue #462, opened 2019-09-05
Classifiers should support gaps in class labels


twotabbies:
## Expected behaviour

Often users are benerating training data based on new observations covering existing already-classified areas. For example, there exist global datasets of land cover classification (e.g., MODIS-derived classification https://developers.google.com/earth-engine/datasets/catalog/MODIS_051_MCD12Q1) and users may be interested in training on new observations, but in areas where some land cover types are not present (e.g., training on farmland in Iowa; there are no glaciers in Iowa). Smile currently doesn't allow this: there must be at least one instance of each class in the training data. There is also no allowance for structural gaps in the class label (again using MCD12Q1.051 as an example, "unclassified" is label 254). While it would be possible for users to map class labels before passing the data to Smile, this is error-prone.

I have prepared a PR that implements this mapping internally to Smile for most classifiers. The ones I haven't implemented it for are LDA, QDA and RDA as their public interface involves a "priori" array, and handling that in the presence of index remapping gets tricky.

## Actual behaviour

Smile's classifiers reject training sets with missing class labels.

haifengl:
v2 has this feature now

PR #463, opened 2019-09-05
Loosen restriction on class labels for most classifiers so that gaps are allowed.

twotabbies:
This addresses #462.

### DIFF
--- a/core/src/main/java/smile/classification/AdaBoost.java
+++ b/core/src/main/java/smile/classification/AdaBoost.java
@@ -88,6 +88,10 @@ public class AdaBoost implements SoftClassifier<double[]> {
      * importance measure.
      */
     private double[] importance;
+    /**
+     * A map from original class labels to the internal dense labels.
+     */
+    private final SparseClassMap labelMap;
 
     /**
      * Trainer for AdaBoost classifiers.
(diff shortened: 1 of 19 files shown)


twotabbies:
Note that this includes the changes in #457.

twotabbies:
Ping - is there any blocker for merging this? This is the biggest missing feature in Smile that's blocking us from rolling it out to the users of Google Earth Engine.

haifengl:
Thanks for contribution. I just don't feel that it is necessary to change all classifiers to support sparse labels. With the utility class, the user can decide if they need to call them before the training and after the prediction. It is better to keep these two parts decoupled and keep the classifier classes simple and clean. Thanks again.

twotabbies:
I feel it's poor API design to require users to rely on wrapper classes to get functionality that's not present in the base classes. It means that N people have to do work rather than 1 person doing work - every new user needs to rediscover the wrapper classes, write code to use them, discover where the seams show through (e.g., if you use a wrapper class for DecisionTree, the class numbers in the result of DecisionTree.dot() won't match the user's classes... and it's basically impossible to patch those up after the fact), and so on. The approach I took affects very little code in the classifier classes: in fact, most classifiers ended up *smaller* after the change.

The need for this is real - I have scientists who are waiting on this feature, and this feature alone, before they can start using the new Smile-based features of Earth Engine. Without this, we'll have to explore other options.

haifengl:
Will the users use smile directly in Earth Engine? Not some layer built on top of it?

twotabbies:
There are a lot of layers in between Smile and the code the users write - so there are opportunities to insert wrappers on our side and so to hide the lack of support for missing class instances. But this wouldn't actually fix the issue completely: as noted, the class labels in the output of .dot() (which would be available to users) would be incorrect. More importantly, so would the class labels in the output of .rpart() (not yet merged), which means that the output of .rpart() wouldn't be able to fulfil its primary purpose of being an interchange format. Fixing this would involve ugly and unreliable patching of the generated strings...

But the point I'm making is that as things stand, I have to write a wrapper class, which I can do... but then so does the next person, and so does any other user who has a similar issue (which is potentially a lot of people - anyone doing local research using a label set developed for global coverage).

haifengl:
In v2, the data frame provides the freedom to use irregular class labels, such as strings, enums, etc. One may encode sparse (integer) class label with NominalScale to handle this issue.

However, there are still use cases that the users want to use the sparse integer class labels directly. I am designing a new Classifier interface to support it. Will update soon.

twotabbies:
Thanks - it sounds like v2 will have a lot of valuable features.

haifengl:
I have incorporate this in v2. Please check smile.classification.ClassLabel and DecisionTree, RandomForest, AdaBoost, and GradientTreeBoosting. Thanks!

haifengl:
Close as we implement this feature in v2. Thanks

PR #464, opened 2019-09-05
Regression tree optimisation

twotabbies:
This uses some of the speedup techniques recently applied to DecisionTree in #448 to speed up RegressionTree. It also reduces RAM usage during and after training.

### DIFF
--- a/core/src/main/java/smile/regression/RegressionTree.java
+++ b/core/src/main/java/smile/regression/RegressionTree.java
@@ -386,7 +386,7 @@ public class RegressionTree implements Regression<double[]> {
         /**
          * Constructor.
          */
-      public TrainNode(Node node, double[][] x, double[] y, int[] samples, int low, int high) {
+        public TrainNode(Node node, double[][] x, double[] y, int[] samples, int low, int high) {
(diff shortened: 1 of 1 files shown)


myui:
LGTM.

haifengl:
Thanks @twotabbies and @myui . I have completely redesigned decision tree and regression tree in v2 branch, which includes many of your ideas. I am still testing it but it should be ready for you to review. The base code is in smile.base.cart. RegressionTree is more mature. Please start there.

twotabbies:
The new RegressionTree, and the CART base class, look very nice; as I was making this PR's changes to RegressionTree, I was wishing that RegressionTree and DecisionTree had a common superclass... and soon they will.

Is there an approximate timeframe for Smile v2? How much more work should be done on v1 while v2 is still being developed?

haifengl:
v2 RC should be ready in a couple of weeks. Since v2 API is very different from v1, I hope that the development on v1 is minimized. Ideally, bug fix only.

haifengl:
I will close this one and let's focus on v2 implementation. Thanks!

PR #465, opened 2019-09-05
Make random forest training deterministic.

twotabbies:
This PR makes each tree in the random forest use a deterministically seeded RNG, using different seeds for each tree. It does this for classification and regression random forests. It also updates the FAQ answer for this question. The run-time cost of this change should be quite low, as no locks are needed, just a single RNG re-seed per tree.

### DIFF
--- a/core/src/main/java/smile/classification/RandomForest.java
+++ b/core/src/main/java/smile/classification/RandomForest.java
@@ -318,6 +318,11 @@ public class RandomForest implements SoftClassifier<double[]> {
          * The out-of-bag predictions.
          */
         int[][] prediction;
+        /**
+         * The RNG seed to use when running the task. Used to make sure that training is
+         * deterministic, even when multi-threading.
+         */
+        int rngSeed;
 
         /**
          * Constructor.
@@ -334,6 +339,7 @@ public class RandomForest implements SoftClassifier<double[]> {
             this.classWeight = classWeight;
             this.order = order;
             this.prediction = prediction;
+            this.rngSeed = Math.randomInt(Integer.MAX_VALUE);
         }
 
         @Override
@@ -341,6 +347,7 @@ public class RandomForest implements SoftClassifier<double[]> {
             int n = x.length;
             int k = smile.math.Math.max(y) + 1;
             int[] samples = new int[n];
+            Math.setSeed(rngSeed);
 
             // Stratified sampling in case class is unbalanced.
             // That is, we sample each class separately.
--- a/core/src/main/java/smile/regression/RandomForest.java
+++ b/core/src/main/java/smile/regression/RandomForest.java
@@ -276,6 +276,11 @@ public class RandomForest implements Regression<double[]> {
          * Out-of-bag sample
          */
         int[] oob;
+        /**
+         * The RNG seed to use when running the task. Used to make sure that training is
+         * deterministic, even when multi-threading.
+         */
+        int rngSeed;
 
         /**
          * Constructor.
@@ -292,12 +297,14 @@ public class RandomForest implements Regression<double[]> {
             this.subsample = subsample;
             this.prediction = prediction;
             this.oob = oob;
+            this.rngSeed = Math.randomInt(Integer.MAX_VALUE);
         }
 
         @Override
         public RegressionTree call() {
             int n = x.length;
             int[] samples = new int[n];
+            Math.setSeed(rngSeed);
 
             if (subsample == 1.0) {
                 // Training samples draw with replacement.
--- a/core/src/test/java/smile/classification/RandomForestTest.java
+++ b/core/src/test/java/smile/classification/RandomForestTest.java
@@ -225,4 +225,26 @@ public class RandomForestTest {
             System.err.println(ex);
         }
     }
+
+    /**
+     * Tests that RNG seeding consistently controls the trainig.
+     */
+    @Test
+    public void testSeeding() throws Exception {
+        System.out.println("Seeding");
+        ArffParser arffParser = new ArffParser();
+        arffParser.setResponseIndex(4);
+        AttributeDataset iris = arffParser.parse(smile.data.parser.IOUtils.getTestDataFile("weka/iris.arff"));
+        double[][] x = iris.toArray(new double[iris.size()][]);
+        int[] y = iris.toArray(new int[iris.size()]);
+
+        Math.setSeed(123);
+        RandomForest forest1 = new RandomForest(iris.attributes(), x, y, 100);
+        Math.setSeed(123);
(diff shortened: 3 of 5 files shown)


myui:
@twotabbies [Math.permutate(variables)](https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/classification/DecisionTree.java#L523) also uses ThreadLocal random. FYI

```java
    public static void permutate(int[] x) {
        random.get().permutate(x);
    }
```

Better to avoid random behavior in utility class/methods.

twotabbies:
This one's tricky no matter how you slice it. One traditional approach involves passing around a PRNG that's explicitly used as a source for randomness during training. That doesn't work well together with multi-threaded training, though - you'd have to make sure that the different threads always interleave their random number requests identically, which isn't feasible, or else hoist all random number generation to the top-level thread ("thread X, here are your pre-generated random numbers. Use them wisely") which requires a lot of restructuring. So you end up needing thread-local PRNGs, and have the master thread control their seeding. This could be done with explicit PRNGs that get passed around, or via the thread-local "global variable" PRNGs that's Smile's current style (e.g., implicitly used by Math.permutate() as you point out). Just controlling the seeding of those PRNGs was the simplest (least code change) approach, so it's the one I went with.

myui:
Math utility class is used among libraries, not only RandomForest and DecisionTrees. Then, fixing a seed impacts on other algorithms using RNGs in Math class.

twotabbies:
I agree that global RNGs are problematic, but they're what we currently have. Any randomised algorithm *ought to* support some kind of seeding, but not enough do; this PR adds it to the place I currently care about :-)

Or is your point that the seeding I do here affects other algorithms that are run on those threads after RandomForest training completes? Yes, it does - but so would any other change that affected RF's use of the RNG, including using more (or fewer) random numbers... those algorithms that don't provide their own seeds should have no expectations about the RNG state when they start running.

myui:
> Or is your point that the seeding I do here affects other algorithms that are run on those threads after RandomForest training completes? 

Yes. Also, enabling [a random seed in DecisionTree](https://github.com/apache/incubator-hivemall/blob/master/core/src/main/java/hivemall/smile/classification/DecisionTree.java#L202) is more preferable IMPO (since DecisionTree can take a random seed).

The workaround in PR can be considered considering the current design. Anyway, I'll leave the decision to @haifengl

haifengl:
I have tried to discourage people to set random number generator seeds because

- People generally don't understand how RNG works and set bad seed.
- Behind the scene, we actually have multiple RNG algorithms and they may be used together in one machine learning algorithm. One seed may not be good for different RNGs.
- As you know, we are multi-threaded and each thread should use a different seed. Otherwise, it doesn't make sense, especially in case of random forest. It is really hard to properly set many seeds in API design and implementation given we don't really know how many threads will be created.
- It loses most of values of a stochastic algorithm if one wants deterministic results!

But I understand that it is needed in case of test. In this case, one can call smile.math.Math.setSeed and also set -Dsmile.threads=1.

twotabbies:
I don't understand what you mean by a "bad seed" - with a good RNG, all seeds should be equally good (or bad). If the RNG has problems with some seeds, that's something that should be fixed at the RNG level. Can you give examples (0? 1?) of seeds that are bad for Smile's current RNGs? (I did notice in passing that the RNGs' actual seed space is quite a lot smaller than it appears to be; setSeet takes a 64-bit long but the actual number of bits used for seeding is under 30.)

I agree that it would be hard to have an API design for random forest training that would allow the user control of individual trees' seeds. I don't think that would be a good idea either. This approach means that users who want deterministic results just call setSeed before training - and a different seed is used for each tree without them having to worry about it.

My approach sidesteps the threading problem entirely by ensuring that each *task* uses a different seed, and that seed will be consistent no matter what thread the task is run on. Multi-threading happens, or not, as the user desires... but the end result is the same. That's a good design for any multi-threaded system.

The real reason for this change is that I very strongly disagree with "It loses most of values of a stochastic algorithm if one wants deterministic results!" - repeatability is an essential part of any large-scale system design. Google has lots of experience with this: on a large enough computer fleet, machine failures (or network failures, or just being evicted by a higher-priority job) happen all the time. When a large job has been divided up into smaller tasks which are run in parallel, you sometimes need to re-run one of those tasks - and in this context it's important to know that the re-run will produce the same results as the original run. It's more complicated than it sounds; the original run may have produced partial results, and if the second run produces a different sequence of results then downstream processing can become corrupted. There's a brief discussion of this in the original Map/Reduce paper at https://static.googleusercontent.com/media/research.google.com/en//archive/mapreduce-osdi04.pdf but there are issues beyond the ones discussed in that paper. A lot of engineering goes into ensuring that any computation is a repeatable function of its inputs... stochastic computations very definitely included.

myui:
+1 to have setSeed feature in DecisionTree because without fixing a random seed, hyperparameter search (e.g., max_depth and min_samples_split) cannot be properly done.

haifengl:
A "bad seed" is a seed containing too little information. 0 and 1 are good examples as you mentioned. I am not aware of a RNG that works well for any seed.

haifengl:
About ""It loses most of values of a stochastic algorithm if one wants deterministic results!", we are talking about different things. I fully agree with you on the case you mentioned. Any deterministic algorithm should produce the same result not matter it is local, parallel or distributed.

But I was talking about stochastic algorithms. They suppose not deterministic.

haifengl:
In case of random forest, the seed is not that critical as we generate only limited number of integers. But for algorithms that need a lot of (real-valued) random numbers, the quality of seed is quite important. For example, Monte Carlo simulations in finance and physics.

haifengl:
@myui if you fix the seed for hyper-parameter searching, you just find the best parameter for that seed.

haifengl:
I understand that we need a neat way to set seed in multi-threading environment. I hope that we can spend more time to design a universal approach for this problem, not just for a particular algorithm.

myui:
I think providing a seed to the first RNG and the first RNG create a seed for other RNGs are enough for Decision Tree.
https://scikit-learn.org/stable/modules/generated/sklearn.tree.DecisionTreeClassifier.html#sklearn.tree.DecisionTreeClassifier

haifengl:
in v2, we ensure to use distinct and deterministic seeds for each thread. See smile.math.MathEx line 116. Note that we don't use nextInt to generate seeds for next thread to avoid possible correlations. We also use (large) probably prime number to avoid bad seeds that have low entropy. Users still can set custom seed with setSeed().  

However, it is not sufficient to algorithm random forest because it is impossible to control which thread in the pool train which trees. Therefore, we allow the user to provide a seed generator/stream to seed each tree. Since seed generation may be time consuming, this is only optional.

haifengl:
Let's close this one as we achieve it in v2. Thanks!

Issue #477, opened 2019-10-31
RidgeRegression fails to compute p-value due to floating point comparison error


wiltseb:
## Expected behaviour

Train a RidgeRegression model on my data, where 
```
X = {
                { -1.6186802663701534E7, 1.5394863704190075E9 },
                { 1.8990838567327976E8, 1.4302105368610876E9 },
                { 3.96003574010262E8, 1.3209347033031676E9 },
                { 6.020987623472443E8, 1.2116588697452476E9 },
                { 8.081939506842256E8, 1.1023830361873276E9 },
                { 1.0142891390212078E9, 9.931072026294076E8 },
                { 1.22038432735819E9, 8.838313690714877E8 },
                { 1.4264795156951709E9, 7.745555355135677E8 },
       }

y = { -1.9367312273661723E9, -2.0487128915354834E9, -2.124363122479608E9, -2.162488870865301E9, -2.162488870865301E9, -2.1243631224796085E9, -2.0487128915354834E9, -1.9367312273661723E9 }

lambda = 3.0
```

## Actual behaviour

The algorithm throws an exception. Here is the stack trace...
```
java.lang.IllegalArgumentException: Invalid x: 1.000000000000001

	at smile.math.special.Beta.regularizedIncompleteBetaFunction(Beta.java:58)
	at smile.regression.RidgeRegressionTest.testy(RidgeRegressionTest.java:214)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
[...]

PR #478, opened 2019-10-31
Fix floating point comparison error in Beta#regularizedIncompleteBetaFunction

wiltseb:
My team discovered https://github.com/haifengl/smile/issues/477#issue-515181409 with our randomized testing.

This PR includes tests that demonstrate the bug (both with the original data in RidgeRegression and the computation in Beta#regularizedIncompleteBetaFunction), a fix for the floating point comparison logic, and simplifies the conditional branching in Beta#regularizedIncompleteBetaFunction.

### DIFF
--- a/math/src/main/java/smile/math/special/Beta.java
+++ b/math/src/main/java/smile/math/special/Beta.java
@@ -54,15 +54,16 @@ public class Beta {
      * Continued Fraction approximation (see Numerical recipies for details)
      */
     public static double regularizedIncompleteBetaFunction(double alpha, double beta, double x) {
-        if (x < 0.0 || x > 1.0) {
+        double EPS = 1e-8;
+        if (x + EPS < 0.0 || x - EPS > 1.0) {
             throw new IllegalArgumentException("Invalid x: " + x);
         }
 
         double ibeta = 0.0;
-        if (x == 0.0) {
+        if (Math.abs(x - 0.0) < EPS) {
             ibeta = 0.0;
         } else {
-            if (x == 1.0) {
+            if (Math.abs(x - 1.0) < EPS) {
                 ibeta = 1.0;
             } else {
                 // Term before continued fraction
--- a/math/src/main/java/smile/math/special/Beta.java
+++ b/math/src/main/java/smile/math/special/Beta.java
@@ -59,23 +59,22 @@ public class Beta {
             throw new IllegalArgumentException("Invalid x: " + x);
         }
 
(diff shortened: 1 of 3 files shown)


math/src/main/java/smile/math/special/Beta.java
```diff
@@ -54,27 +54,27 @@ public static double beta(double x, double y) {
      * Continued Fraction approximation (see Numerical recipies for details)
      */
     public static double regularizedIncompleteBetaFunction(double alpha, double beta, double x) {
-        if (x < 0.0 || x > 1.0) {
+        double EPS = 1e-8;
+        if (x + EPS < 0.0 || x - EPS > 1.0) {
```
haifengl: This is incorrect. x should be in (0, 1). This change will allow invalid x mathematically.

haifengl:
1. The change is incorrect in mathematics
2. In v2, you can turn off p-value calculation.
3. If there is issue to calculate p-value, the model parameter is better to be adjusted rather than forcefully modifying fundamental math function.

math/src/main/java/smile/math/special/Beta.java
wiltseb: It only allows values that are 1e-8 less than zero or 1e-8 greater than 1, to account for floating point arithmetic errors. It's a standard practice to account for such errors when comparing floating point numbers: https://stackoverflow.com/questions/1088216/whats-wrong-with-using-to-compare-floats-in-java

The range of possible values for x is inclusive of 0 and 1, right? The check here is for < and >, not <= and >=; plus the x == 0.0 and x==1.0 conditions are handled below...

Now that I've thought about it more, this is a more straightforward implementation:
```
        if (Math.abs(x - 0.0) < EPS) {
            return 0.0;
        }

        if (Math.abs(x - 1.0) < EPS) {
            return 1.0;
        }

        if (x < 0.0 || x > 1.0) {
            throw new IllegalArgumentException("Invalid x: " + x);
        }
```
If after reviewing my reasoning, you'd like to make the change and you prefer the latter method, I can push another commit.

haifengl: Yes, it is inclusive and B(0,a,b)=0 and B(1,a,b)=1 always. The latter is much better. Thanks

haifengl: I will add this to v2. no need more PR. Thanks

haifengl:
The fix is v2 branch now. Thanks

PR #479, opened 2019-10-31
Fix error message in KNN related to missing class

kuzmichp:


haifengl:
Thanks. We have completely changes the handling of class labels in v2.

kuzmichp:
Thanks for pointing that out.

PR #494, opened 2019-12-02
remove unused import which might cause compilation error

rayeaster:
since JDK11, javafx is no longer default shipped with jdk, so this unused import might cause issue when using latest version of JDK


haifengl:
I have optimized many imports. There are in master. Thanks.

Issue #502, opened 2020-01-03
Support reading CSV from other locations


benmccann:
## Expected behaviour

I'd like to be able to read a CSV from over the network, from a `String`, etc. A method that takes an `InputStream` or `Reader` would be great

## Actual behaviour

Only accepts `Path`

## Information

- Smile 2.0

haifengl:
We support reading data from HDFS/S3/HTTP etc. Checkout master. It is not CSV but all other file types.

PR #503, opened 2020-01-03
DataFrame creation from CSV Reader

benmccann:
Closes https://github.com/haifengl/smile/issues/502

I think that `charset` would better be taken as a method parameter rather than a constructor parameter because if you supply a `Reader` it already has an associated `charset` and thus ignores the one from the constructor. However, I didn't change this in order to maintain backwards compatibility. Let me know if you'd like me to change

### DIFF
--- a/io/src/main/java/smile/io/CSV.java
+++ b/io/src/main/java/smile/io/CSV.java
@@ -17,9 +17,14 @@
 
 package smile.io;
 
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
 import java.io.IOException;
+import java.io.StringReader;
+import java.io.Reader;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


io/src/main/java/smile/io/CSV.java
```diff
@@ -100,17 +103,26 @@ public DataFrame read(Path path) throws IOException {
...
             // infer the schema from top 1000 rows.
-            schema = inferSchema(path, Math.min(1000, limit));
+            schema = inferSchema(reader, Math.min(1000, limit));
         }
 
         StructField[] fields = schema.fields();
         List<Function<String, Object>> parser = schema.parser();
 
-        try (CSVParser reader = CSVParser.parse(path, charset, format)) {
+        try (CSVParser csvParser = CSVParser.parse(reader, format)) {
```
haifengl: This is troublesome. inferSchema already read some data, which will be missed by CSVParser here. It is why we pass a Path instead of InputStream/Reader.

benmccann: Good call out. I added a call to `reader.reset()` to address. ot all readers are able to be reset, but it should still allow this method to be called with a wider range of inputs

haifengl:
Thanks. I prefer a complete correct solution. If the Reader implementation doesn't support reset and doesn't throw exception, the user will lose some data silently. It will be very hard for them to figure out why.

What remote data do you want to read? Http?

benmccann:
If a `Reader` does not support `reset` then it will throw an exception. `java.io.IOException: reset() not supported`. We can explicitly check for it with `markSupported()` if you'd like, but I don't think it's really necessary since `reset` will already throw an exception if not supported

haifengl:
For "complete correct", I mean that it should not only work with some Reader implementations. If the goal is to read remote files, it is better to pass `URL` rather than `Reader`. And for v2, I care the API design very much. I would like a good solution which is applicable to other data format too.

benmccann:
Ok, I've made it work for all `Reader` implementations and added a test as well

`URL` to me is less complete. It means you can't accept data from `String` and many other data types. Also, reading from a `URL` may involve dealing with connection timeouts, etc. and I think the user will have more control over those aspects if we accept a `Reader` that they are able to create. However, if you'd like me to add a utility method accepting a `URL` I'd be happy to do so

benmccann:
I probably don't need this anymore. I was trying to convert a TableSaw Table into Smile DataFrame. I was thinking about going through an in-memory CSV as a quick implementation since both libraries support CSV, but I think I figured out how to create the Smile DataFrame directly. That being said, I do think there's still value in accepting input sources other than File, so will leave this open.

haifengl:
Not sure how do you convert between tablesaw and smile now. In-memory CSV is definitely slow though. You may wrap tablesaw row into smile Tuple and then create the DataFrame from a list of tuples.

We will definitely support other input than files. But I don't think InputStream/Reader is good in the API. Not all data format can be processed in stream way.

benmccann:
> But I don't think InputStream/Reader is good in the API. Not all data format can be processed in stream way.

The Apache CSV parser requires that the input be able to be converted to a `Reader`. I'm not sure what other types you're thinking of, but if it can't be converted to a `Reader` I'm not sure how you would process that input

haifengl:
I don't mean just CSV files but all data/file formats supported by smile (e.g. Parquet, Avro, SAS, ...)

benmccann:
Ok, well I will close this for now

Issue #515, opened 2020-02-05
it would be intersting to add tree-SHAP to smile


rayeaster:
SHAP is a recently proposed algorithm for interpretable machine learning by 

> Lundberg, Scott M., and Su-In Lee. “A unified approach to interpreting model predictions.” Advances in Neural Information Processing Systems. 2017

and tree-SHAP is a specific version of implementation for tree-based machine learning like CART, Random Forest, Gradient Boost, etc.

> Lundberg, Scott M., Gabriel G. Erion, and Su-In Lee. “Consistent individualized feature attribution for tree ensembles.” arXiv preprint arXiv:1802.03888 (2018)

> https://github.com/slundberg/shap/blob/master/shap/explainers/pytree.py

rayeaster:
implementation ready for review:  https://github.com/haifengl/smile/pull/524

rayeaster:
implementation review https://github.com/haifengl/smile/pull/530

haifengl:
I have merged shap branch into master. Thanks!

PR #524, opened 2020-03-16
add tree shap importance

rayeaster:
implementation for https://github.com/haifengl/smile/issues/515

example shap value summary plot for UCI Cervical Cancer data using https://github.com/jtablesaw/tablesaw:

[](https://postimg.cc/ygtcGRYM)


io/src/main/java/smile/io/CSV.java
```diff
@@ -147,10 +147,13 @@ private DataFrame read(Reader reader, int limit) throws IOException {
             for (CSVRecord record : csv) {
                 Object[] row = new Object[fields.length];
                 for (int i = 0; i < fields.length; i++) {
-                    String s = record.get(i).trim();
-                    if (!s.isEmpty()) {
-                        row[i] = parser.get(i).apply(s);
-                    }
+					String recordI = record.get(i);
+					if (recordI != null) {
```
haifengl: Why is it possible null?

core/src/main/java/smile/regression/TreeShapImportance.java
```diff
@@ -0,0 +1,735 @@
+package smile.regression;
```
haifengl: Can you please format the code by our standard? IntelliJ has shortcut to do so. Thanks.

io/src/main/java/smile/io/CSV.java
rayeaster: CSV convert missing value to null.

core/src/main/java/smile/regression/TreeShapImportance.java
rayeaster: I use eclipse, will try the code format feature in Eclipse.  Do you have the code format from IntelliJ to import, so I might could use it in Eclipse to format the code? Thanks

core/src/test/java/smile/regression/GradientTreeBoostTest.java
```diff
@@ -64,6 +66,9 @@ public void testLongley() throws Exception {
         MathEx.setSeed(19650218); // to get repeatable results.
         GradientTreeBoost model = GradientTreeBoost.fit(Longley.formula, Longley.data);
 
+        TreeShapExplainer shapExplainer = new TreeShapExplainer(model.trees(), true);
```
haifengl: This is NOT real test. Can you get the result from other implementation and compare to them here?

core/src/main/java/smile/regression/TreeShapImportance.java
```diff
@@ -0,0 +1,743 @@
...
+					trees.length);
+			for (int i = 0; i < trees.length; i++) {
+				ConcurrentShapConstruction shapRunnable = new ConcurrentShapConstruction(trees[i], concurrencySentinel,
+						this.normalize);
+				exec.submit(shapRunnable);
+				concurrencyRunnables.add(shapRunnable);
+			}
+
+			try {
+				System.out.println("----- waiting for all shap trees get created -----");
```
haifengl: please use logger if it really outputs something important.

core/src/main/java/smile/regression/TreeShapImportance.java
```diff
@@ -0,0 +1,743 @@
...
+		 * will be normalized)
+		 */
+		public TreeShapExplainer(RegressionTree[] trees, boolean normalize) {
+			this.normalize = normalize;
+			this.trees = new ShapTree[trees.length];
+
+			final CountDownLatch concurrencySentinel = new CountDownLatch(trees.length);
+			final ExecutorService exec = Executors.newFixedThreadPool(Math.max(threadNum, 1));
+
+			List<ConcurrentShapConstruction> concurrencyRunnables = new ArrayList<ConcurrentShapConstruction>(
```
haifengl: This can be done by Stream.parallel().

core/src/main/java/smile/regression/TreeShapImportance.java
```diff
@@ -0,0 +1,743 @@
...
+ *
+ * <p>
+ *
+ * @author ray
+ * @see
+ *     <pre>
+ * https://christophm.github.io/interpretable-ml-book/shap.html
+ * </pre>
+ */
+public class TreeShapImportance {
```
haifengl: It is not necessary to have two nested classes.

core/src/main/java/smile/regression/TreeShapImportance.java
```diff
@@ -0,0 +1,743 @@
...
+		void setValue(double value) {
+			this.value = value;
+		}
+
+		double getValue() {
+			return this.value;
+		}
+	}
+
+	private static class MutableInteger {
```
haifengl: we have smile.util.MutableInt

core/src/test/java/smile/regression/GradientTreeBoostTest.java
rayeaster: Thanks for review. will get implementation comparison here

rayeaster:
@haifengl  update the PR with suggest change. but not ready to final view. pending on an issue as https://github.com/haifengl/smile/issues/526

core/src/main/java/smile/base/cart/CART.java
```diff
@@ -271,12 +271,12 @@ protected boolean split(final Split split, PriorityQueue<Split> queue) {
         int mid = split.lo + trueSamples.length;
 
         LeafNode trueChild = newNode(trueSamples);
-        assert trueChild.size == split.trueCount : String.format("trueChild.size != split.trueCount: %d != %d", trueChild.size, split.trueCount);
-        assert trueChild.size >= nodeSize : String.format("trueChild size is too small: %d < %d", trueChild.size, nodeSize);
+        //assert trueChild.size == split.trueCount : String.format("trueChild.size != split.trueCount: %d != %d", trueChild.size, split.trueCount);
```
rayeaster: this is not intended change, but pending on resolution for https://github.com/haifengl/smile/issues/526

io/src/main/java/smile/io/CSV.java
haifengl: This is not necessary. We handle missing value correctly. You got issue here because you specify wrong data type. For double with missing values, it should be `DoubleObject`. BTW, you don't need specify the schema in most type as we automatically infer csv schema.

core/src/test/java/smile/data/NHANES.java
```diff
@@ -0,0 +1,88 @@
...
+
+  public static DataFrame data;
+  public static Formula formula = Formula.lhs("y");
+
+  public static double[][] x;
+  public static double[] y;
+
+  static {
+    List<StructField> fields = new ArrayList<StructField>();
+    fields.add(new StructField("y", DataTypes.DoubleType));
```
haifengl: not necessary. Read.csv will infer schema.

core/src/main/java/smile/regression/treeshap/MutableDouble.java
```diff
@@ -0,0 +1,14 @@
+package smile.regression.treeshap;
```
haifengl: This should be in the same package of MutableInt. The API should be consistent.

haifengl: Add license at the head of all files.

haifengl:
Generally, we would like to minimize the number of modified files for a feature. This single feature changes 16 files, which is hard to track and may easily break other things.

core/src/test/java/smile/data/NHANES.java
rayeaster: > not necessary. Read.csv will infer schema.

inferred schema will have generated field name like V1/V2/...? thought it would be more straight-forward to add schema explictly in this dataset specific class.

rayeaster:
> Generally, we would like to minimize the number of modified files for a feature. This single feature changes 16 files, which is hard to track and may easily break other things.

actually not too many existing code are touched except one line in CSV to allow ignore header. will remove changes in other tests (regression/randomforest/gradienttreeboost) f that helps.

core/src/test/java/smile/data/NHANES.java
haifengl: The header line will be used for variable names.

core/src/main/java/smile/regression/treeshap/ShapCalculationTemp.java
```diff
@@ -0,0 +1,62 @@
...
+    this.s = s;
+    this.feature_indexes = new ArrayList<MutableInt>(s);
+    this.zero_fractions = new ArrayList<MutableDouble>(s);
+    this.one_fractions = new ArrayList<MutableDouble>(s);
+    this.pweights = new ArrayList<MutableDouble>(s);
+
+    returnTemp(true);
+  }
+
+  void returnTemp(boolean init) {
```
haifengl: I don't feel that it is necessary to create this class if all we need is this function.

core/src/main/java/smile/regression/treeshap/ShapTree.java
```diff
@@ -0,0 +1,151 @@
...
+  public int max_depth;
+
+  public ShapTree(RegressionTree tree) {
+    this(tree, false);
+  }
+
+  public ShapTree(RegressionTree tree, boolean normalize) {
+    this.normalize = normalize;
+
+    List<Integer> childrenLefts = new ArrayList<Integer>();
```
haifengl: Please avoid List of primary types if possible.

io/src/main/java/smile/io/CSV.java
```diff
@@ -144,13 +144,19 @@ private DataFrame read(Reader reader, int limit) throws IOException {
 
         try (CSVParser csv = CSVParser.parse(reader, format)) {
             List<Tuple> rows = new ArrayList<>();
-            for (CSVRecord record : csv) {
+            List<CSVRecord> records = csv.getRecords();
```
haifengl: Why still modify CSV.java?

rayeaster: > Why still modify CSV.java?

will remove this modification.

core/src/main/java/smile/regression/treeshap/ShapCalculationTemp.java
```diff
@@ -0,0 +1,62 @@
...
+ * <pre>
+ * https://github.com/slundberg/shap/blob/master/shap/explainers/pytree.py
+ * </pre>
+ * 
+ * @author ray
+ */
+public class ShapCalculationTemp {
+
+  public int s;
+  public List<MutableInt> feature_indexes;
```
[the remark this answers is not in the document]
rayeaster: These lists are already get pre-allocated with proper size estimate and never need re-creation on the fly. 

But since Java do not have a convenient and performance-friendly sub-range support for plain array type, while List class do have such support via `subList` view which is preferred here to avoid complex implementation code (otherwise we might need manually handle the subrange indexing into the plain array and modify many functions accordingly in the algorithm to accept subrange indexes as we traverse down the tree)

rayeaster:
> Generally, we would like to minimize the number of modified files for a feature. This single feature changes 16 files, which is hard to track and may easily break other things.

PR updated with respect to suggest review. reduce the files to 10 from 16, including 3 new dataset files, 3 new dataset classes, 1 new test class and 3 implementation related classes.

Thanks again for taking time to review

core/src/test/java/smile/data/BostonHousing.java
```diff
@@ -0,0 +1,47 @@
+/**
```
haifengl: Please fix the header. Similarly for other files. Thanks

core/src/main/java/smile/regression/treeshap/ShapTree.java
```diff
@@ -0,0 +1,164 @@
...
+  public int[] children_left;
+  public int[] children_right;
+  public int[] children_default;
+  public int[] features;
+  public Node[] thresholds;
+  public double[] values;
+  public double[] node_sample_weight; // equals number of samples for a node divided by total number of samples
+  public int max_depth;
+
+  public ShapTree(RegressionTree tree) {
```
haifengl: It looks to me that this class can be applied to DecisionTree too. Shall we take `smile.base.cart.CART`? If so, how about moving this class to `smile.feature` package?

core/src/main/java/smile/regression/treeshap/ShapTree.java
```diff
@@ -0,0 +1,164 @@
...
+    } 
+
+    // we recompute the expectations to make sure they follow the SHAP logic
+    this.max_depth = computeExpectations(this.children_left, this.children_right, this.node_sample_weight, this.values, 0, 0);
+    // tracking progress for shap tree construction 
+    System.out.print(".");
+  }
+
+  /** calculate the constant prediction value of each node */
+  private double subtreeValues(Node n) {
```
haifengl: What do we need if it is a decision tree?

rayeaster: As far as I understand, shap algorithm need the output from leaf node so that is why I only add RegressionTree in constructor. Not sure what we gonna do with DecisionTree since the leaf node could only generate a class label instead of some real-value output.

you may find more details at the original paper, page 3 to page 4
https://arxiv.org/pdf/1802.03888.pdf

core/src/test/java/smile/data/BostonHousing.java
rayeaster: I do not add particular code to handle headers manually. just use `Read.arff` as you suggest. Do I miss anything?

haifengl: The license header of source code file is ill-formated.

core/src/main/java/smile/regression/treeshap/ShapTree.java
```diff
@@ -0,0 +1,164 @@
...
+/** 
+ * structure for shap value calculation. implementation reference:
+ * 
+ * <pre>
+ * https://github.com/slundberg/shap/blob/master/shap/explainers/pytree.py
+ * </pre>
+ * 
+ * @author ray
+ */
+public class ShapTree {
```
haifengl: Please rename to `TreeSHAP`.

core/src/main/java/smile/regression/treeshap/ShapTree.java
haifengl: Even though it is only applicable to regression tree, I still think `smile.feature` is a better home. No need to create a subpackage `treeshap`.

core/src/main/java/smile/regression/treeshap/TreeShapImportance.java
```diff
@@ -0,0 +1,523 @@
...
+ *
+ * <p>
+ *
+ * @author ray
+ * @see
+ *     <pre>
+ * https://christophm.github.io/interpretable-ml-book/shap.html
+ * </pre>
+ */
+public class TreeShapImportance {
```
haifengl: Please rename to `EnsembleTreeSHAP`

core/src/main/java/smile/feature/EnsembleTreeSHAP.java
```diff
@@ -0,0 +1,531 @@
...
+	 * <p>
+	 * This is useful if you want to have deep insight about effect of each feature
+	 * over prediction by using returned shap value for every data instance to draw
+	 * a summary plot between each feature and prediction
+	 * 
+	 * @param x data {@link Tuple}
+	 */
+	public double[] shapValues(Tuple x) {
+		int s = maxd * (maxd + 1);
+		List<MutableInt> feature_indexes = initMutableList(s, false);
```
haifengl, line 141: We don't need to use List and MutableDouble at all. Why not just double[]?

core/src/main/java/smile/feature/TreeSHAP.java
```diff
@@ -0,0 +1,164 @@
...
+ * 
+ * <pre>
+ * https://github.com/slundberg/shap/blob/master/shap/explainers/pytree.py
+ * </pre>
+ * 
+ * @author ray
+ */
+public class TreeSHAP {
+	
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(TreeSHAP.class);
```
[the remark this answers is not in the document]
rayeaster: do you have a handy format plugin for eclipse you would recommend?

haifengl:
Can you please remove new test data set? As long as the test on Boston Housing data can pass the test, it is good enough. The data sets in the repo is already very large. I prefer not introducing more unless necessary.

core/src/main/java/smile/feature/TreeSHAP.java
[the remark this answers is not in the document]
haifengl: You don't need plugin. You can configure the formatting.

haifengl:
Please change the target branch to a feature branch (says shap). It is not a good idea to PR into master directly.

rayeaster:
> Please change the target branch to a feature branch (says shap). It is not a good idea to PR into master directly.

Would you please create a feature branch in your repo named shap so that I could push request to. I will close this PR by now. Thanks

Issue #526, opened 2020-03-22
smile.regression.RandomForest::predict() sometimes returns NaN for very few samples


tuxdna:
## Expected behaviour

I train a RandomForest regression model on my data which has ~ 3000 samples and 8 attributes, all attributes are double precision, with the predicted value being a double in the range 0 to 1 in the training dataset. After traning, when we make predict() call on all the samples from training data, sometimes one or two samples result in NaN, instead of a valid double prediction.

## Actual behaviour

```
Found NaN
sampleNumber: 582
trueValue: 1.0
predictedValue: NaN
trueValue - predictedValue: NaN 

features: [17.0, 3.0, 0.058823529411764705, 0.7647058823529411, 1.5176745294625686, 0.19293206793206796, 1.0, 1.0]

```

## Code snippet

```java
        int numTrees = 25;
        int nodeSize = 1;
        int maxDepth = 20;
        double sampleRate = 0.85;

        Properties props = new Properties();
        props.setProperty("smile.random.forest.trees", numTrees + "");
        props.setProperty("smile.random.forest.node.size", nodeSize + "");
        props.setProperty("smile.random.forest.max.depth", maxDepth + "");
        int maxLeafNodes = trainDataSet.size() / 3;
        props.setProperty("smile.random.forest.max.nodes", maxLeafNodes + "");
        props.setProperty("smile.random.forest.sample.rate", sampleRate + "");
        RandomForest rf = RandomForest.fit(formula, trainDataSet, props);
```

## Input data

The data is described above. Can't share the actual dataset.

## Information

- Java 

```
openjdk version "1.8.0_242"
OpenJDK Runtime Environment (build 1.8.0_242-8u242-b08-0ubuntu3~18.04-b08)
OpenJDK 64-Bit Server VM (build 25.242-b08, mixed mode)
```

- Smile version 2.2.2
-  Ubuntu 18.04 64bit

tuxdna:
On debugging the code, of all the RegressionTree instances in the forest ( obtained via tf.trees() ), one RegressionTree predict()'s a NaN value. This apparently happens because the path traversed on this tree doesn't correspond to any samples from training data i.e. the LeafNode -- new RegressionNode(n, out, mean, rss) will mean of NaN.

So a quick fix could be to just ignore this prediction and return  a mean of remaining RegressionTree's predictions:

```diff
diff --git a/core/src/main/java/smile/regression/RandomForest.java b/core/src/main/java/smile/regression/RandomForest.java
[...]

PR #527, opened 2020-03-22
Fix issue #526 - RandomForest::predict() sometimes returns NaN

tuxdna:
Potential fix for: https://github.com/haifengl/smile/issues/526


rayeaster:
does the data contains NaN or missing value?

tuxdna:
@rayeaster The training has no missing or NaN values.

PR #534, opened 2020-04-08
tune the refactored unwindPathSum

rayeaster:
seems the parameter for `GradientTreeBoost` affect the shap value for this dataset but also a minor calculation issue found in `unwindPathSum` function.

another minor concern is the default for-loop in `TreeSHAP` iterating over all trees might cause a lot double/int array to create along the path if we have thousands of trees (could easily double memory footprint from 200MB to 400MB for some big dataset like `ailerons` with 5000 trees). Since the tree calculation is sequentially/independently one by one, some kind of memory optimization might be a choice there? for example, we could re-use (reset after use) same pre-allocated array among trees (so the array created on the fly would be like `O(1)` instead of `O(# of trees)`), but that require sub-indexing support in `extendPath`,`unwindPath`,`unwindPathSum` and `recurse`. 

just my 2 cents. Thanks

### DIFF
--- a/core/src/main/java/smile/regression/RegressionTree.java
+++ b/core/src/main/java/smile/regression/RegressionTree.java
@@ -493,17 +493,18 @@ public class RegressionTree extends CART implements Regression<Tuple>, DataFrame
             double n = w[l];
             if (po != 0) {
                 for (int j = l - 1; j >= 0; j--) {
-                    double t =  n / ((j+1) * po);
+                    double t =  (n * (l + 1)) / ((j+1) * po);
                     sum += t;
-                    n = w[j] - t * pz * (l - j);
-                }
+                    n = w[j] - t * pz * ((l - j) / (l + 1));
+                }                
+                return sum;
             } else {
                 for (int j = l - 1; j >= 0; j--) {
                     sum += w[j] / (pz * (l - j));
                 }
+                return sum * (l + 1);
             }
-
-            return sum * (l + 1);
+            
         }
     }
 }
--- a/core/src/test/java/smile/regression/GradientTreeBoostTest.java
+++ b/core/src/test/java/smile/regression/GradientTreeBoostTest.java
@@ -159,7 +159,7 @@ public class GradientTreeBoostTest {
     @Test
     public void testShap() {
         MathEx.setSeed(19650218); // to get repeatable results.
-        GradientTreeBoost model = GradientTreeBoost.fit(BostonHousing.formula, BostonHousing.data, Loss.ls(), 100, 6, 100, 5, 0.05, 0.7);
+        GradientTreeBoost model = GradientTreeBoost.fit(BostonHousing.formula, BostonHousing.data, Loss.ls(), 100, 3, 10, 5, 0.01, 1);
         double[] importance = model.importance();
         double[] shap = model.shap(BostonHousing.data.stream().parallel());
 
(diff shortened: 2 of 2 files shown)


core/src/main/java/smile/regression/RegressionTree.java
```diff
@@ -493,17 +493,18 @@ void unwind(int i) {
             double n = w[l];
             if (po != 0) {
                 for (int j = l - 1; j >= 0; j--) {
-                    double t =  n / ((j+1) * po);
+                    double t =  (n * (l + 1)) / ((j+1) * po);
```
haifengl, line 496: I refer the c++ version: https://github.com/slundberg/shap/blob/cde70feb211443cfbd9f599fcded8fe00b9d6a3d/shap/tree_shap.h#L377-L390.

Why is it wrong?

core/src/test/java/smile/regression/GradientTreeBoostTest.java
```diff
@@ -159,7 +159,7 @@ public void testHuber() {
     @Test
     public void testShap() {
         MathEx.setSeed(19650218); // to get repeatable results.
-        GradientTreeBoost model = GradientTreeBoost.fit(BostonHousing.formula, BostonHousing.data, Loss.ls(), 100, 6, 100, 5, 0.05, 0.7);
+        GradientTreeBoost model = GradientTreeBoost.fit(BostonHousing.formula, BostonHousing.data, Loss.ls(), 100, 3, 10, 5, 0.01, 1);
```
haifengl, line 162: subsmple = 1.0 doesn't make much sense for gradient tree boosting. It is why I change it to 0.7. maxDepth = 3 is also too shallow to test some cases. Can you get the values for my settings?

core/src/main/java/smile/regression/RegressionTree.java
rayeaster: ok... maybe the c++ version is a bit different from the python version: https://github.com/slundberg/shap/blob/master/shap/explainers/pytree.py#L291-L302

core/src/test/java/smile/regression/GradientTreeBoostTest.java
rayeaster: ignoring above change in `unwindPathSum`, no other obvious calculation issue by me for the implementation. the test shap number is as below for your parameter setting:
```
----- SHAP -----
ZN              0.0052
CHAS            0.0067
INDUS           0.0294
RAD             0.0316
B               0.0482
TAX             0.0807
CRIM            0.0994
AGE             0.1206
NOX             0.1341
PTRATIO         0.1530
DIS             0.1581
RM              0.5589
LSTAT           0.9871
```

haifengl: I double checked unwindPathSum and also verify the math formula. I think that my implementation is correct (also faster).

Thanks for the new SHAP values. These are much closer to my numbers. But why the numbers in the older PR are much bigger?

rayeaster: > I double checked unwindPathSum and also verify the math formula. I think that my implementation is correct (also faster).
> 
> Thanks for the new SHAP values. These are much closer to my numbers. But why the numbers in the older PR are much bigger?

different GradientTreeBoost parameter make the majority of difference IMO. not the shap calculation.

rayeaster:
I assume I could close this PR?

haifengl:
Yes, we can close this one. Thanks.

PR #558, opened 2020-06-11
minor fix to logging message in UMAP code

rgwood-github2018dec31:
I don't see a 'canary' br so hoping this is ok...

### DIFF
--- a/core/src/main/java/smile/manifold/UMAP.java
+++ b/core/src/main/java/smile/manifold/UMAP.java
@@ -582,7 +582,7 @@ public class UMAP implements Serializable {
                 }
             }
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/smile/manifold/UMAP.java
```diff
@@ -582,7 +582,7 @@ private static void optimizeLayout(double[][] embedding, double[] curve, SparseM
                 }
             }
 
-            logger.info(String.format("The learning rate at %3d iterations: %.5f", iter, alpha));
+            logger.info(String.format("The learning rate at %3d iterations: %.5f", alpha, iter));
```
haifengl, line 585: This is wrong. Note that the first string interpolation is `%3d` for iterations.

PR #562, opened 2020-06-16
Trim must be done after generating spaces

tomassvensson:


haifengl:
Thanks. Can you provide an example that breaks the current implementation?

tomassvensson:
Sure, here is one in Kotlin:

        assertEquals("X",  smile.nlp.normalizer.SimpleNormalizer.getInstance().normalize("\u00A0X"))

where \u00A0 is a non-breaking space.

haifengl:
Thanks!

PR #563, opened 2020-06-21
Polish BarPlot legends and data

albertmolinermrf:
### Description
`smile.plot.swing.BarPlot` allows setting legends but currently they are not displayed.

There is also a small typo in an array index which leads to odd plots if `smile.plot.swing.BarPlot#of(double[][], java.lang.String[])` is used.


haifengl:
Thanks!

PR #568, opened 2020-07-03
initial commit to add ARIMA

rayeaster:
added

- ARIMA(p,d,q) implementation 
- 3 timeseries datasets
- tests and demo

sample stats chart on female birth data:

image


core/src/test/java/smile/data/DailyFemaleBirth.java
```diff
@@ -0,0 +1,73 @@
...
+ * @author rayeaster
+ */
+public class DailyFemaleBirth {
+
+    public static DataFrame data;
+    public static double[] timeseries;
+
+    private static final String DATA_COL_NAME = "births";
+    static {
+        ArrayList<StructField> fields = new ArrayList<>();
```
haifengl, line 53: Have you tried parsing without schema? Wonder if read.csv can infer the schema correctly.

math/src/main/java/smile/math/matrix/Matrix.java
```diff
@@ -456,6 +456,32 @@ default SVD svd(int k, double kappa, int maxIter) {
             return new SVD(U, V, s);
         }
     }
+    
+    /**
```
haifengl, line 460: I have redesigned the matrix library. When it is ready, I will merge it to arima branch.

core/src/test/java/smile/data/BitcoinHistoryPrice.java
```diff
@@ -0,0 +1,84 @@
...
+ *   "Feb 18, 2018",11123.40,11349.80,10326.00,10551.80,"8,744,010,000","187,663,000,000"
+ *   ...
+ * </pre>
+ * 
+ * @see https://www.kaggle.com/sudalairajkumar/cryptocurrencypricehistory
+ * @see https://arxiv.org/pdf/1904.05315.pdf
+ * 
+ * @author rayeaster
+ */
+public class BitcoinHistoryPrice {
```
haifengl, line 48: Let's just call it BitcoinPrice. We don't have future price anyway :)

math/src/main/java/smile/math/matrix/Matrix.java
```diff
@@ -456,6 +456,32 @@ default SVD svd(int k, double kappa, int maxIter) {
...
+     * @return a {@link DenseMatrix} of Toplietz characteristic
+     */
+    static DenseMatrix toeplitz(double[] input) {
+        int length = input.length;
+        double toeplitz[][] = new double[length][length];
+
+        for (int i = 0; i < length; i++) {
+            for (int j = 0; j < length; j++) {
+                if (j > i) {
+                    toeplitz[i][j] = input[j - i];
```
haifengl, line 475: So this is also a symmetric matrix?

rayeaster: I believe yes, it is symmetric. https://en.wikipedia.org/wiki/Toeplitz_matrix

core/src/test/java/smile/data/DailyFemaleBirth.java
rayeaster: let me try later

math/src/main/java/smile/math/matrix/Matrix.java
haifengl: I have merged arima with master. Please update. Thanks

rayeaster: @haifengl I updated the PR. However, with new change around `Matrix` and its related methods like `Cholesky`, the `ARIMATest` which could pass before now break. I tried several ways but not working as expected. I thought the new change to `Matrix` be transparent to the caller but it require mkl libraray now as well. Before it was some pure Java implementation like `JMatrix`.

Issue #491, opened 2019-12-01
KDTree should use distance squared, not distance


lukehutch:
`KDTree.java` currently uses distance, not distance squared. This wastes work, since the `sqrt` is not necessary.

https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/neighbor/KDTree.java#L238
https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/neighbor/KDTree.java#L280
https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/neighbor/KDTree.java#L324

haifengl:
I used to use MathEx.squaredDistance and do `sqrt` on the final result. In v2, neighbor fields become `public final` so that we have to calculate the distance in the first place. For real life data, we should not have notable performance difference for this small change.

lukehutch:
Well, for my real world use case, I need to create hundreds of thousands of k-d trees, and perform trends of thousands of tests on each one... So every wasted cycle does matter!

haifengl:
How about you do performance comparison with your real case and let's know the difference between distance and squaredDistance?

lukehutch:
Sorry for the delay in benchmarking this. I finally got around to doing a distance comparison.

I made a copy of the `KDTree` class, and switched the code to used `MathEx.squaredDistance` rather than `MathEx.distance`.

Results for 1.2M queries of a three-dimensional `KDTree` containing 40000 points, running on a single core of a Ryzen-1800X CPU, on JDK 1.8:

* `distance`: 17738.37 queries/sec (67.65 sec total)
* `squaredDistance`: 139448.25 queries/sec (8.61 sec total)
* Speedup: 7.86x

Your assumption that `For real life data, we should not have notable performance difference for this small change` is incorrect.

Can you please reopen this bug?

haifengl:
Thanks. Have you run your changes with unit tests?

lukehutch:
I have unit tests for my own project, and those tests pass identically for the squared distance version. I didn't create a PR against the git latest version, because of your comment about fields being made `public static`. The changes are very straightforward, but I don't know what you want to do about the fact that you already publicly exposed the `distance` field in the class. Please advise.

haifengl:
skipping sqrt cannot make 7.8x difference. I think that you miss the change to KD-Tree branch logic. The unit tests has a case on it. Please try it.

lukehutch:
[...]

PR #571, opened 2020-07-05
Switch distance to distanceSq

lukehutch:
EXPERIMENTAL PULL REQUEST ONLY -- do not merge.

Replaces `distance` with `distanceSq` (see #491).

Contains just enough changes to have `sbt 'testOnly **.KDTreeTest'` pass all unit tests.

Benchmark code:

```java
    public static void main(String[] args) {
        int N = 40000;
        int size = 100;
        int numTests = 100_000;
        
        Random rand = new Random(1);
        
        double[][] coords = new double[N][3];
        Integer[] vals = new Integer[N];
        for (int i = 0; i < N; i++) {
            coords[i][0] = rand.nextDouble() * size - size/2;
            coords[i][1] = rand.nextDouble() * size - size/2;
            coords[i][2] = rand.nextDouble() * size - size/2;
            vals[i] = i;
        }
        KDTree<Integer> kdt = new KDTree<>(coords, vals);
        
        long startTime = System.nanoTime();
        double[] q = new double[3];
        for (int i = 0; i < numTests; i++) {
            q[0] = rand.nextDouble() * size - size/2;
            q[1] = rand.nextDouble() * size - size/2;
            q[2] = rand.nextDouble() * size - size/2;
            Neighbor<double[], Integer> nearest = kdt.nearest(q);
            // System.out.println(nearest.index);
        }
        System.out.println("Took " + (System.nanoTime() - startTime) * 1.0e-9 + " sec");
    }
```

Uncomment the `println` command to verify that the same exact nearest neighbors are returned for the original (git latest) version of SMILE and this modified version.

Benchmark: 

* Original: 13.74 sec
* This modified version: 0.11 sec
* Speedup: 124x

### DIFF
--- a/core/src/main/java/smile/manifold/NearestNeighborGraph.java
+++ b/core/src/main/java/smile/manifold/NearestNeighborGraph.java
@@ -93,7 +93,7 @@ class NearestNeighborGraph {
                 int v1 = i;
                 for (int j = 0; j < neighbors.length; j++) {
                     int v2 = neighbors[j].index;
-                    double weight = neighbors[j].distance;
+                    double weight = Math.sqrt(neighbors[j].distanceSq);
                     graph.setWeight(v1, v2, weight);
                     consumer.accept(v1, v2, weight, j);
                 }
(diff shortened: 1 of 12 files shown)


haifengl:
Thanks. It is really not about sqrt. There is a bug at Line 269 and 313. When I changed to distance, I should have to compare to abs(diff). When you switch back to squaredDistance, you also switch back to diff * diff, which accidentally fix the bug.

lukehutch:
Ah, that makes a lot more sense! Thanks for taking a look and finding the bug.

Issue #575, opened 2020-07-08
Line based dataframe


JD557:
**Is your feature request related to a problem? Please describe.**

With the new smile 2.0 came a new `DataFrame` whose default implementation (`DataFrameImpl`) is column-based.

While this sounds like a good idea in theory, I'm hitting some performance problems in practice (e.g. a workload that took 5s to train now takes 4min to convert the data into a `DataFrame` and 1s to train).

**Describe the solution you'd like**

I think this could be solved by having a line-based implementation of a dataframe. While training models might be slower, this could be a nice tradeoff in some cases.

**Describe alternatives you've considered**

I tried to implement this myself by attempting to implement something like a `MyBaseVector(columnIdx: Int)` that would work as a view over the line-based structure.
I think there are only two problematic methods here:

* `Object array();` - There's not an actual array backing the `BaseVector`. This could be implemented by copying everything to an array when this is called, but that could have some weird performance implications.
* `BaseVector<T, TS, S> get(int... index)`/`BaseVector<T, TS, S> get(int... index)` - This would require storing the filtered indices as well (like `MyBaseVector(columnIdx: Int, filteredEntries: Option[Set[Int]])` or with an extra `FilteredVector(innerVector: BaseVector, filteredEntries: Set[Int])`). I guess it's possible, assuming that the list of indices is small.

None of those methods seems to be used in the DataFrame, so that should be OK.

haifengl:
How do you convert data to DataFrame?

JD557:
I'm currently doing something like `DataFrame.of(trainingFeatures.asJava)`, where `trainingFeatures` is a scala `List[Tuple]` that was read via `akka-streams` from a `gRPC` endpoint (so I think that I can't use anything from `smile.io` to help me here).

haifengl:
`akka-stream` is lazy. `DataFrame.of(trainingFeatures.asJava)` in fact includes the stream processing. Why don't you measure the time in two steps:
```
val data = trainingFeatures.asJava
time {
  DataFrame.of(data)
}
```
`DataFrame.of()` shouldn't use 4 minutes.

JD557:
OK, I think that I found the problem:

The `DataFrameImpl` constructor receives a `List`, which it then uses on a lot of lines like `for (int i = 0; i < size; i++) values[i] = (int) read.invoke(data.get(i));`.

I was using a `LinkedList`, so that `data.get(i)` was really expensive. Converting everything to a vector beforehand fixed the problem.

[...]

PR #576, opened 2020-07-08
Stop using random access in the DataFrameImpl constructor

JD557:
Related to #575 

The `DataFrameImpl` constructor was using random access when it was reading the data in a sequential fashion. This could lead to an accidentally quadratic complexity if the `data` was backed by a linked list or a similar data structure.

Since the access was sequential anyway, using a `for(T value: data)` loop and incrementing a separate `i` variable should do the trick without any performance loss.


haifengl:
Thanks. I already fixed it in v2 branch. BTW, two constructors need to be fixed.

JD557:
Nice 👍

Issue #604, opened 2020-10-26
Optimization for radius-limited search in KDTree


lukehutch:
In the KDTree `buildNode` method, the spread of the node is calculated:

```
        // Calculate bounding box stats
        double maxRadius = -1;
        for (int i = 0; i < d; i++) {
            double radius = (upperBound[i] - lowerBound[i]) / 2;
            if (radius > maxRadius) {
                maxRadius = radius;
                node.split = i;
                node.cutoff = (upperBound[i] + lowerBound[i]) / 2;
            }
        }
```

However, maxRadius is never saved to the node, i.e. the `Node` class could have a `double maxRadius` field, and in the innermost block of the code above, then you can modify the line in the inner block:

```
                node.maxRadius = maxRadius = radius;
```

If this `maxRadius` information were stored in each node, then all the radius-limited search methods could terminate early if *neither* child could contain a point within the range, by inserting the following check before recursing on either child:

```
    if (Math.abs(diff) > node.maxRadius + radius) {
        return;
    }
```

haifengl:
How much speed up do you observe? Where (line #) exactly do you add the check? Do it pass all the unit tests? Thanks.

lukehutch:
All tests pass with the change. However I noticed some strange behavior in some cases, and I realized that it's because kd-tree search works bottom-up, not top-down. I'll close this for now, but will submit another PR if I can find a better way to prune the search tree.

PR #606, opened 2020-10-26
Optimization for range-limited KDTree search (#604)

lukehutch:
See #604. All tests pass.


lukehutch:
See #604, this doesn't work in the current form.

Issue #607, opened 2020-10-28
Incompatibility of One Versus Rest and Random Forest


damienlancry:
**Is your feature request related to a problem? Please describe.**
the current code for OneVersusRest seems to be incompatible with Random Forest. Indeed all [fit methods](https://github.com/haifengl/smile/blob/a1720b837e2e00b1cb1122b1e0882ef78e81bd92/core/src/main/java/smile/classification/RandomForest.java#L273) of the class Random Forest take a Formula and a DataFrame as inputs. 
In contrast the OneVersusRest Classifier [fit method](https://github.com/haifengl/smile/blob/a1720b837e2e00b1cb1122b1e0882ef78e81bd92/core/src/main/java/smile/classification/OneVersusRest.java#L89) takes features x: Array[T], labels y: Array[Int] and a BiFunction trainer as input. As illustrated in the documentation:
```
    val zip = read.csv("data/usps/zip.train", delimiter = ' ' , header = false)
    val x = zip.drop(0).toArray
    val y = zip.column(0).toIntArray
    val kernel = new GaussianKernel(8.0)
    val model = ovr(x, y) { (x, y) => SVM.fit(x, y, kernel, 5, 1E-3) }
```

**Describe the solution you'd like**
A solution that would be nice would be to have a fit method for the Random Forest that takes features x: Array[T], and labels y: Array[Int] so that we could train a one versus rest random forest like that:
```
    val model = ovr(x, y) { (x, y) => RandomForest.fit(x, y) }
```

**Describe alternatives you've considered** 
Another solution that could work would be to have a fit method for the OneVersusRest class that takes a Formula and a DataFrame as input so that we could train a one versus rest random forest like that:
```
    val model = ovr(formula, data) { (formula, data) => RandomForest.fit(formula, data) }
```

**Additional context**
There could have been a work around to that by doing something like :
```
val model = ovr(x, y) { (x, y) => RandomForest.fit("class" ~, DataFrame.of(x).merge(DataFrame.of(y.map(Array(_)), "class")))}
```
But this gives the error:
```
<console>:46: error: type mismatch;
 found   : smile.classification.RandomForest
 required: smile.classification.Classifier[Array[Double]]
```
Another challenge that will arise is that by default the one versus rest performs platt scaling on the output of the f method of its base estimator and the f method is only implemented for some classifiers (svms, logistic regressions, max entropy...).
Therefore, one would need to make platt scaling an option, and use the predict method of the classifiers instead of f.

[...]

PR #611, opened 2020-10-29
[WIP] fix incompatibility between one versus rest and DataFrameClassifiers

damienlancry:
### Your checklist for this pull request
🚨 Please review the [guidelines for contributing](../CONTRIBUTING.md) to this repository.

- [ x] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!
- [x ] Make sure you are making a pull request against the **canary branch** (left side). Also you should start *your branch* off *our canary*.
- [ x] Check the commit's or even all commits' message styles matches our requested structure.
- [x ] Check your code additions will fail neither code linting checks nor unit test.

### Description
Please describe your pull request.

Hi, I developed a fix to make OneVersusRest Compatible with SoftClassifier. This is related to [this issue](https://github.com/haifengl/smile/issues/607)
To do so, I defined a new fit method for OneVersusRest.
I also modified the predict so that it uses platt scaling only for non SoftClassifiers.
It is now possible to do that:
```
val data = read.csv("data/usps/zip.train", delimiter = ' ' , header = false)
val formula = "V1" ~
val model = ovr(formula, data) { (formula, data) => RandomForest.fit(formula, data) }
```
I also changed a tiny bit the CrossValidation to make it compatible with this:
```
cv.classification(10, formula, data) { case (formula, data) => ovr(formula, data) { (formula, data) => RandomForest.fit(formula, data) } }
```
I put it as WIP because I could add some tests and maybe some scaladoc.
I would love to have your feedback.

Thank you!

### DIFF
--- a/core/src/main/java/smile/classification/DataFrameClassifier.java
+++ b/core/src/main/java/smile/classification/DataFrameClassifier.java
@@ -27,7 +27,7 @@ import smile.data.type.StructType;
  *
  * @author Haifeng Li
  */
-public interface DataFrameClassifier {
+public interface DataFrameClassifier extends SoftClassifier<Tuple> {
 
     /**
      * Predicts the class label of an instance.
--- a/core/src/main/java/smile/classification/OneVersusRest.java
+++ b/core/src/main/java/smile/classification/OneVersusRest.java
@@ -18,6 +18,10 @@
 package smile.classification;
 
 import java.util.function.BiFunction;
+import smile.data.DataFrame;
+import smile.data.formula.Formula;
+import smile.data.Tuple;
+import smile.data.vector.BaseVector;
 import smile.math.MathEx;
 import smile.util.IntSet;
 
@@ -60,11 +64,16 @@ public class OneVersusRest<T> implements SoftClassifier<T> {
      * Constructor.
      * @param classifiers the binary classifier for each one-vs-rest case.
      */
+    public OneVersusRest(Classifier<T>[] classifiers) {
+        this(classifiers, null);
+    }
+
+    /**
+     * Constructor.
+     * @param classifiers the binary classifier for each one-vs-rest case.
+     */
     public OneVersusRest(Classifier<T>[] classifiers, PlattScaling[] platts) {
-        this.classifiers = classifiers;
-        this.platts = platts;
-        k = classifiers.length;
-        labels = IntSet.of(k);
+        this(classifiers, platts, IntSet.of(classifiers.length));
     }
 
     /**
@@ -107,7 +116,7 @@ public class OneVersusRest<T> implements SoftClassifier<T> {
         ClassLabels codec = ClassLabels.fit(y);
         int k = codec.k;
         if (k <= 2) {
-            throw new IllegalArgumentException(String.format("Only %d classes" + k));
+            throw new IllegalArgumentException("Only %d classes" + k);
         }
 
         int n = x.length;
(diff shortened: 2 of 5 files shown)


core/src/main/java/smile/classification/OneVersusRest.java
```diff
@@ -107,7 +116,7 @@ public OneVersusRest(Classifier<T>[] classifiers, PlattScaling[] platts, IntSet
         ClassLabels codec = ClassLabels.fit(y);
         int k = codec.k;
         if (k <= 2) {
-            throw new IllegalArgumentException(String.format("Only %d classes" + k));
+            throw new IllegalArgumentException("Only %d classes" + k);
```
damienlancry, line 119: oops here it should be 
```
            throw new IllegalArgumentException(String.format("Only %d classes", k));
```

haifengl: Please fix it.

core/src/main/java/smile/classification/OneVersusRest.java
```diff
@@ -128,12 +137,58 @@ public OneVersusRest(Classifier<T>[] classifiers, PlattScaling[] platts, IntSet
         return new OneVersusRest<>(classifiers, platts);
     }
 
+    /**
```
haifengl, line 140: It is better not to duplicate the code for Tuple/DataFrame. I would rather to check if the classifier support `f(x)` function (by catching the exception). If not, don't fit platt scaling.

core/src/main/java/smile/validation/CrossValidation.java
```diff
@@ -137,11 +139,11 @@ public CrossValidation(int n, int k, boolean permutate) {
      * Runs cross validation tests.
      * @return the predictions.
      */
-    public int[] classification(Formula formula, DataFrame data, BiFunction<Formula, DataFrame, DataFrameClassifier> trainer) {
+    public int[] classification(Formula formula, DataFrame data, BiFunction<Formula, DataFrame, SoftClassifier<Tuple>> trainer) {
```
haifengl, line 142: It may not be a `SoftClassifier`.

haifengl:
I push some changes to OvO and OvR to master. Please incorporate. Thanks.

core/src/main/java/smile/classification/OneVersusRest.java
```diff
@@ -138,12 +150,58 @@ public OneVersusRest(Classifier<T>[] classifiers, PlattScaling[] platts, IntSet
...
+        return new OneVersusRest<>(classifiers);
+    }
+
     @Override
     public int predict(T x) {
         int y = 0;
         double maxf = Double.NEGATIVE_INFINITY;
         for (int i = 0; i < k; i++) {
-            double f = platts[i].scale(classifiers[i].f(x));
+            double f = platts != null? platts[i].scale(classifiers[i].f(x)) : classifiers[i].predict(x);
```
damienlancry, line 204: here this part is trickier than I thought because I assumed classifiers[i].predict(x) would return a double that would be the predicted probability, when actually it returns a int that is either 0 or 1 in the one versus rest strategy. so here the predicted class is the first class that is positively predicted. for instance if we have 3 classes and we train 3 classifiers with the one versus rest strategy, say the first classifier predict probability of .6 for the first class, the second classifier predict probability of .3 for the second class, and the third classifier predict probability of .9 for the third class, the predicted class would be the first class here, when it should be the third one. 
In my tests, I still got 92% accuracy on usps dataset with one versus rest random forest (balanced, so I tricked the classWeight parameter to balance the classes by default), which mislead me to think this implementation was correct.

core/src/main/java/smile/classification/OneVersusRest.java
damienlancry: I agree there is a lot of duplicated code and I do not like it either. But the main reason I defined a new fit method is because the types do not match with the previous fit method: the previous fit takes parameters (T[] x, int[] y, int pos, int neg, BiFunction<T[], int[], Classifier<T>> trainer) and there is no simple way to apply it on a DataFrame that is why I defined a new fit method with parameters (Formula formula, DataFrame data, int pos, int neg, BiFunction<Formula, DataFrame, SoftClassifier<Tuple>> trainer).

But if we simply add a check for support for f function, then the workaround I proposed in the issue #607 should work:
```
val xt = x.map(arr => Tuple.of(arr, zip.drop(0).schema()))
val model = ovr(xt, y) { (xt, y) => RandomForest.fit("class" ~, DataFrame.of(xt.map(_.toArray)).merge(DataFrame.of(y.map(Array(_)), "class")))}
```
(Indeed it works, but I did not test the accuracy because cv fails:
```
cv.classification(10, xt, y) { case (xt, y) => ovr(xt, y) { (xt, y) => RandomForest.fit("class" ~, DataFrame.of(xt.map(_.toArray)).merge(DataFrame.of(y.map(Array(_)), "class"))) } }
```
fails with the error:
```
<console>:49: error: value toArray is not a member of (Double, Double, Double, Double, Double, Double, Double, Double, Double, Double, Double, Double, Double, Double, Double, Double, Double, Double, Double, Double, Double, Double)
       cv.classification(10, xt, y) { case (xt, y) => ovr(xt, y) { (xt, y) => RandomForest.fit("class" ~, DataFrame.of(xt.map(_.toArray)).merge(DataFrame.of(y.map(Array(_)), "class"))) } }
```
)
But I am worried about the efficiency of converting array of tuples  x to array of arrays of doubles, to dataframe and merging with dataframe of labels converted from array of ints y, and all of this k times (k being the number of classes).

core/src/main/java/smile/validation/CrossValidation.java
damienlancry: Yes I wanted to leave DataFrameClassifier in the first place, but I had some type errors because the DataFrameClassifier interface do not extend the Classifier interface and thus the classifiers attribute of the OneVersusRest cannot be a list of DataFrameClassifier s. In contrast, the SoftClassifier interface extends the Classifier interface so it is possible for the classifiers attribute of the OneVersusRest class to be an array of SoftClassifier s. Moreover, to the best of my knowledge, all classifiers that implements DataFrameClassifier also implements SoftClassifier, that is why I thought this would not be an issue. But I might be wrong on that last point.

core/src/main/java/smile/classification/OneVersusRest.java
```diff
@@ -160,7 +218,7 @@ public int predict(T x, double[] posteriori) {
         }
 
         for (int i = 0; i < k; i++) {
-            posteriori[i] = platts[i].scale(classifiers[i].f(x));
+            posteriori[i] = platts != null? platts[i].scale(classifiers[i].f(x)) : classifiers[i].predict(x);
```
damienlancry, line 221: same problem would arise here (if there was no error thrown as you added in master).
indeed the posteriori array would be an array of zeros and ones. the whichMax utility function will choose the first one, so the predictions will be the same as above.

core/src/main/java/smile/classification/DataFrameClassifier.java
```diff
@@ -27,7 +27,7 @@
  *
  * @author Haifeng Li
  */
-public interface DataFrameClassifier {
+public interface DataFrameClassifier extends SoftClassifier<Tuple> {
```
haifengl, line 30: This is wrong. They are orthogonal.

core/src/main/java/smile/classification/OneVersusRest.java
```diff
@@ -57,6 +61,14 @@
     /** The class label encoder. */
     private IntSet labels;
 
+    /**
+     * Constructor.
+     * @param classifiers the binary classifier for each one-vs-rest case.
+     */
+    public OneVersusRest(Classifier<T>[] classifiers) {
```
haifengl, line 68: no need this constructor.

core/src/main/java/smile/classification/DataFrameClassifier.java
damienlancry: ok I will remove it but then I am stuck I am afraid because of the problem above

core/src/main/java/smile/classification/OneVersusRest.java
```diff
@@ -155,13 +225,17 @@ public int predict(T x) {
...
-        }
-
-        for (int i = 0; i < k; i++) {
-            posteriori[i] = platts[i].scale(classifiers[i].f(x));
-        }
+            if (classifiers[0] instanceof SoftClassifier) {
+                for (int i = 0; i < k; i++) {
+                    double[] proba = new double[2];
+                    ((SoftClassifier<T>) classifiers[i]).predict(x, proba);
+                    posteriori[i] = proba[1];
```
haifengl, line 232: This is wrong. posterior probability is not defined this. Your sum is eve not 1.

haifengl:
I add a new fit method that takes DataFrame and returns DataFrameClassifier.

PR #655, opened 2021-04-12
Fixed bug in "ones"

darrenjw:
This is a single line change (5 character insertion) to fix a very obvious bug in "ones" when the matrix is not square. It currently returns a matrix of zeros rather than a matrix of ones in that case.


haifengl:
Thanks!

PR #667, opened 2021-05-24
Fix scala DataFrameOps partition

ahmed-mahran:
`l :+ i` returns a copy of the original collection with `i` appended, hence `l` and `r` are always empty
`l += i` appends `i` to the original collection, hence `l` and `r` will contain the required partitioning


haifengl:
Thanks!

Issue #671, opened 2021-06-10
Coursier Bootstrapping not Building Almond Kernel


ASRagab:
**Describe the bug**
When attempting to run the `./shell/src/universal/bin/jupyterlab.sh` to setup smile for jupyterlab according to the quickstart guide for the project under _Notebooks_

I received the following error:
```bash
Resolution error: Error downloading sh.almond:scala-kernel_2.13.5:0.11.2
  not found: /Users/ahmadragab/.ivy2/local/sh.almond/scala-kernel_2.13.5/0.11.2/ivys/ivy.xml
  not found: https://repo1.maven.org/maven2/sh/almond/scala-kernel_2.13.5/0.11.2/scala-kernel_2.13.5-0.11.2.pom
  not found: https://jitpack.io/sh/almond/scala-kernel_2.13.5/0.11.2/scala-kernel_2.13.5-0.11.2.pom
./shell/src/universal/bin/jupyterlab.sh: line 67: ./almond-scala-2.13: No such file or directory
```

Using `master` branch

The issue is that the almond jars are not available for `2.13.5` as of yet, specifically the `scala-kernel-api` jar is only published for `2.13.4`.  Αdditionally, current versions of coursier seem to now require a main class be declared when building the launcher I will submit a PR, that successfully installed the kernel.

PR #672, opened 2021-06-10
Coursier Bootstrapping #671

ASRagab:
Issue #671

- setting scala version back to 2.13.4
- setting `main-class`
- Updating github.io website to note location of script `took second to find` 

Kernel is installed and working

Screen Shot 2021-06-10 at 1 35 57 AM

### DIFF
--- a/shell/src/universal/bin/jupyterlab.sh
+++ b/shell/src/universal/bin/jupyterlab.sh
@@ -53,7 +53,7 @@ install_almond() {
       chmod +x coursier
   fi
 
-  SCALA_VERSION=2.13.4 ALMOND_VERSION=0.11.2
+  SCALA_VERSION=2.13.5 ALMOND_VERSION=0.11.2
 
   ./coursier bootstrap \
       -r jitpack \
--- a/shell/src/universal/bin/jupyterlab.sh
+++ b/shell/src/universal/bin/jupyterlab.sh
@@ -53,7 +53,7 @@ install_almond() {
       chmod +x coursier
   fi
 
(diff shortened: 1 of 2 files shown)


shell/src/universal/bin/jupyterlab.sh
```diff
@@ -53,7 +53,7 @@ install_almond() {
       chmod +x coursier
   fi
 
-  SCALA_VERSION=2.13.5 ALMOND_VERSION=0.11.2
```
haifengl, line 56: No need to roll back. When we will make a new release, it should be fine. Just keep change in your own fork.

web/quickstart.html
```diff
@@ -415,7 +415,8 @@ <h2 id="notebook">Notebooks</h2>
     <p>You can also use Smile in your favorite Notebook.
         We recommend JupyterLab and provide <code>jupyterlab.sh</code>
         to setup the conda environment of Jupyter Lab for Smile with
-        kernels for Scala, Kotlin and Clojure. When you run
+        kernels for Scala, Kotlin and Clojure. The script is located at
+        <code>./shell/src/universal/bin/jupyterlab.sh</code>. When you run
```
haifengl, line 419: The script is in `bin` directory of release package. It is better not post the source code path here.

shell/src/universal/bin/jupyterlab.sh
ASRagab: Makes sense! Wii change.

ASRagab:
@haifengl updated, I look forward to learning more about smile and contributing to this project!

haifengl:
Thanks!

Issue #682, opened 2021-07-10
Negative scores in BM25


schneijan:
**Describe the bug**
It's not really a bug, rather a suggestion for an improvement. Currently, the inverse document frequency (IDF) in BM25 is calculated using `Math.log((N - n + 0.5) / (n + 0.5))`. However, this way negative IDF values are possible, for example when the term occurs in every single document of the corpus. Since the other part of the BM25 weighing scheme is multiplied with the IDF, also the overall BM25 score can become negative. This is typically undesired and the reason why other people recommend to use `Math.log((N - n + 0.5) / (n + 0.5) + 1)` instead for calculating the IDF. [Wikipedia](https://en.wikipedia.org/wiki/Okapi_BM25) states this IDF formula and other relevant NLP platforms like [Lucene seem to use similar approaches](https://opensourceconnections.com/blog/2015/10/16/bm25-the-next-generation-of-lucene-relevation/) as well for calculating BM25. For this reason, I would suggest to change the IDF formula within the BM25 class accordingly.

**Expected behavior**
The calculated BM25 score should always be greater than or equal to zero.

**Actual behavior**
The calculated BM25 score may become negative if the respective term occurs in a high proportion of the documents within the corpus.

**Code snippet**
```
int freq = 3;
int docSize = 100;
int avgDocSize = 150;
int N = 10000000;
int n = 9999900;
double result = new BM25().score(freq, docSize, avgDocSize, N, n);
System.out.printf("BM25 score: %f%n", result);
```
Prints:
> BM25 score: -30,982883

In PR #683 the suggested changes are implemented.

Thanks for having a look!

haifengl:
Merged. Thanks!

PR #683, opened 2021-07-10
Avoid negative BM25 scores

schneijan:
### Description
This PR implements the suggestions of #682 by adjusting the IDF formula within the _BM25.java_ class at three places. Due to this change, the expected value of the corresponding unit test changed slightly and is hence adjusted as well.

Please tell me if something is wrong with this PR or if you do not see a need for this formula change. Thanks in advance!


haifengl:
Thanks!

PR #703, opened 2021-12-21
fix test compilation with transformer

rayeaster:


haifengl:
Thanks. I updated these tests.

PR #706, opened 2022-01-12
fix test data path

rayeaster:
It might help to add `/data` sub-folder in default system property value directly instead using string concatenation later. And plain string concatenation would encounter path seperator issue across platforms like Linux(using `/`) and Windows (using `\\` in Java).

For example, in Windows, with proposed change, developer could run the unit test with java option `-Dsmile.home=C:\smile\shell\src\universal\data` to override the default value

### DIFF
--- a/base/src/main/java/smile/util/Paths.java
+++ b/base/src/main/java/smile/util/Paths.java
@@ -18,6 +18,7 @@
 package smile.util;
 
 import java.io.BufferedReader;
+import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.stream.Stream;
@@ -29,7 +30,7 @@ import java.util.stream.Stream;
  */
 public interface Paths {
     /** Smile home directory. */
-    String home = System.getProperty("smile.home", "shell/src/universal/data");
+    String home = System.getProperty("smile.home", "shell/src/universal/");
 
     /**
      * Get the file path of a test sample dataset.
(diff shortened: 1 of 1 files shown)


haifengl:
Thanks. But it will breaks the semantics of `smile.home` that should point to the top directory of installation. This may cause issues with smile shell where `smile.home` is set in the startup script.

rayeaster:
> Thanks. But it will breaks the semantics of `smile.home` that should point to the top directory of installation. This may cause issues with smile shell where `smile.home` is set in the startup script.

how about introduce another variable specific for testing data like `smile.data` so it could avoid string concatenation and keep existing `smile.home` semantic intact

haifengl:
Try this for line 40

```
return java.nio.file.Paths.get(home + File.separator + "data", path);
```

It should be platform independent.

rayeaster:
> Try this for line 40
> 
> ```
> return java.nio.file.Paths.get(home + File.separator + "data", path);
> ```
> 
> It should be platform independent.

It works and PR updated accordingly

base/src/main/java/smile/util/Paths.java
```diff
@@ -29,15 +30,15 @@
  */
 public interface Paths {
     /** Smile home directory. */
-    String home = System.getProperty("smile.home", "shell/src/universal/data");
```
haifengl: This line should not be changed.

rayeaster: > This line should not be changed.

That line is not included in the final changes contained in this PR
image

haifengl:
Thanks!

PR #857, opened 2026-04-11
fix(FPTree): guard against negative index in freq when all counts are zero

tejasae-afk:
Found a small bug in `FPTree.freq` while going through the codebase.

The `while (f[--n] == 0)` loop has no lower-bound guard. On an empty itemset stream all frequencies stay zero, so `n` keeps decrementing past 0 and hits `f[-1]`, throwing `ArrayIndexOutOfBoundsException`.

Happy to close if this is intentional, you know the codebase far better than I do.


haifengl:
Thanks for reporting edge case. We throw an InvalidArgumentException on empty stream now. It is better to fail fast.

Issue #872, opened 2026-07-18
Smile Studio: notebook cells accumulate a trailing blank line on every save/open


TsybulkaM:
### Describe the bug

Every time a notebook is opened in Smile Studio, an extra blank line is appended to the end of each cell. The blank lines accumulate on every save → reopen cycle, so cells slowly fill up with trailing empty lines.

### Expected behavior

Opening and saving a notebook should be idempotent: the cell contents (including their line count) stay exactly the same across repeated save/open cycles. No blank lines should be added.

### Actual behavior

Trailing blank lines grow by one per reopen. Cause is a mismatch between how cells are written and read in smile.studio.notebook.Notebook:

`saveAsSource()` joins cells with "\n" + separator + "\n" — i.e. it writes a `\n` before each separator.
`getSourceCells()` builds each cell with current.append(line).append('\n') — i.e. it adds a trailing \n to the last line of every cell.
Together, each round-trip leaves one more trailing newline:

```
foo   →  save  →  foo\n//--- CELL ---\n...   →  open  →  cell = "foo\n"
"foo\n" → save → foo\n\n//--- CELL ---\n...  →  open  →  cell = "foo\n\n"
```

### Additional context

Java: OpenJDK (Temurin) 25
Smile version: 6.2.4
Build system / OS: Linux (Ubuntu-family), built with sbt

haifengl:
Thanks for contributing! PR is merged.

PR #873, opened 2026-07-18
Smile Studio: notebook cells accumulate a trailing blank line on every save/open

TsybulkaM:
Description in ticket: https://github.com/haifengl/smile/issues/872


haifengl:
Thanks!

Issue #772, opened 2024-05-08
F1, precision & recall for multi-class classification


sandervh14:
**Is your feature request related to a problem? Please describe.**
I am evaluating a multi-class classification case. The values in the target column and prediction columns can be 0, 1, 2, 3 or 4.
The FScore, Precision and Recall calculation results in the following error:

```
java.lang.IllegalArgumentException: Precision can only be applied to binary classification: 1000

	at smile.validation.metric.Precision.of(Precision.java:54)
	at smile.validation.metric.FScore.of(FScore.java:90)
	at smile.validation.metric.FScore.score(FScore.java:47)
```

**Describe the solution you'd like**
Could you implement this, please? Calculating precision and recall for multi-class classification should be possible.

**Describe alternatives you've considered**
Not applicable.

**Additional context**
Not applicable.

haifengl:
Sure, we will add support for multi class. BTW, `smile.deep.metric` package precision and recall support macro-, micro-, and weighted metrics of multi-class data.

haifengl:
It is added to master branch now. See [tests](https://github.com/haifengl/smile/blob/master/core/src/test/java/smile/validation/metric/PrecisionTest.java) for examples.

sandervh14:
Hi @haifengl,

Thanks for your pointer to `smile.deep.metric` and for implementing on the master branch!
And sorry for replying so late. Due to a higher priority work item added to our backlog, we had postponed the work in which I needed this. Only now I'm getting back to it.

Thanks!
Sander

sandervh14:
Hi @haifengl,

Could you make a new release?
I see on https://github.com/haifengl/smile/blob/master/core/src/main/java/smile/validation/metric/Recall.java that the 3.1.1 release does not yet contain the changes in master.
If not possible yet just now, I'll wait. No problem.

Sander

PR #874, opened 2026-07-21
Fix macro and weighted F-score to average per-class scores

gaoflow:
For multi-class inputs, `FScore.of` with `Averaging.Macro`/`Weighted` returns Fβ of the *averaged* precision and recall instead of the mean of the *per-class* Fβ:

```java
double p = Precision.of(truth, prediction, strategy); // mean of the per-class precision
double r = Recall.of(truth, prediction, strategy);    // mean of the per-class recall
return (1 + beta2) * (p * r) / (beta2 * p + r);        // Fβ of two averages
```

Fβ is not linear in precision/recall, so `Fβ(mean P, mean R)` is not the macro/weighted F-score. The macro precision and macro recall themselves are already correct; only their combination into the F-score is off. The multi-class averaging support was added for #772.

Example — truth `{0,0,1,1,2,2}`, prediction `{0,1,1,1,2,2}` — against scikit-learn `fbeta_score`:

| metric | smile | scikit-learn |
|---|---|---|
| macro precision | 0.888889 | 0.888889 |
| macro recall | 0.833333 | 0.833333 |
| macro F1 | 0.860215 | 0.822222 |
| macro F2 | 0.843882 | 0.821549 |
| macro F0.5 | 0.877193 | 0.849206 |

Weighted is wrong the same way once supports are unequal: for truth `{0,0,0,1,1,2}`, prediction `{0,0,1,1,2,2}`, weighted F1 is 0.705882 vs 0.677778.

Fix: for Macro/Weighted, compute the per-class Fβ from the per-class counts and average them (unweighted for Macro, support-weighted for Weighted), matching `fbeta_score` including its `zero_division=0` for empty classes. Micro and binary are untouched (micro precision == micro recall; binary is a single class).

The two existing macro/weighted assertions pinned the old values and were updated; the new tests check the small examples above against scikit-learn. Ran `sbt "core / testOnly smile.validation.*"` (167 tests, green).

The torch-backed `deep.metric.F1Score` combines its precision and recall the same way; happy to address it in a follow-up.

### DIFF
--- a/core/src/main/java/smile/validation/metric/FScore.java
+++ b/core/src/main/java/smile/validation/metric/FScore.java
@@ -17,6 +17,7 @@
 package smile.validation.metric;
 
 import java.io.Serial;
+import smile.math.MathEx;
 
 /**
  * The F-score (or F-measure) considers both the precision and the recall of the test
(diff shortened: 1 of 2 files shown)


haifengl:
Thank you very much for contributing!

haifengl:
@gaoflow would you please make a PR to fix `deep.metric.F1Score` too? Thanks a lot!

PR #877, opened 2026-07-26
Fix macro/weighted F1 and two sibling averaging bugs in deep.metric

gaoflow:
Follow-up to #874, as requested there.

`deep.metric.F1Score` had the same F-of-means defect that #874 fixed in `smile.validation.metric.FScore`. While auditing the rest of the package against scikit-learn I found two more variants of the same mistake — the reduction of an averaging strategy applied to the wrong quantity, or without the guard the other strategies have — so this fixes the class rather than the one method.

### 1. Macro and weighted F1 averaged the wrong thing

`F1Score.compute()` called `precision.compute()` and `recall.compute()`, which already reduce the per-class vector to a scalar, then took their harmonic mean. F1 is not linear in (P, R), so `F1(mean P, mean R)` is not the mean of the per-class F1 scores.

Confusion matrix (rows = truth, columns = prediction):

```
        pred 0  pred 1  pred 2   support
true 0       4       6       0        10
true 1       0       2       0         2
true 2       0       0       3         3
```

per-class precision = [1, 1/4, 1], recall = [2/5, 1, 1], F1 = [4/7, 2/5, 1]

|             | before     | correct              |
|-------------|------------|----------------------|
| macro F1    | 0.77419355 | 23/35 = 0.65714286   |
| weighted F1 | 0.72       | 111/175 = 0.63428571 |

### 2. Micro recall stopped equalling accuracy when a class is empty

`Recall.compute()` divided the micro score by `where(size > 0, size, 1).sum()`. That guard exists to keep the per-class recall of an empty class at 0 instead of NaN, but summing it makes every empty class add one phantom sample to the micro denominator. `Averaging.Micro`'s javadoc states that micro precision and recall equal accuracy, and `validation.metric.Recall` divides by `n`.

With 3 output classes, no instance of class 2 in the batch and 8 of 10 samples correct, micro recall was 8/11 = 0.72727275 and micro F1 was 16/21 = 0.76190478, against an accuracy of 0.8. Any mini-batch that happens to miss a rare class hits this.

### 3. Weighted metrics returned NaN after reset()

The weighted reduction `score.mul(size).sum().div(size.sum())` has no guard for the empty state, so `Precision`, `Recall` and `F1Score` all returned NaN after `reset()` — called at the end of every epoch, per the `Metric` javadoc. Macro and micro return 0 there, and there is already a test asserting that for macro F1.

### Changes

`Precision.score()` and `Recall.score()` expose the per-class score with the guard each already had, so `F1Score` can average the per-class F1 without duplicating them. `Averaging.weighted()` performs the support-weighted reduction once for the three metrics. `Recall.compute()` now branches on the strategy rather than on `tp.size(0)`.

### Verification

Differential run against scikit-learn 1.8.0 over 10 label sets (perfect, imbalanced 3-class 900/80/20, imbalanced 4-class, binary, never-predicted class, empty class, all-wrong, multi-batch, random 5-class, single-class truth), covering every metric and averaging strategy — 109 comparisons. 18 divergences before, none after, largest remaining difference 9e-08 in float32. Accuracy, all three precisions, macro/weighted recall and the binary metrics already agreed, so the audit closes the class.

7 new tests in `MetricTest`, every one of them failing before the change. `./gradlew :deep:test` goes from 391 tests / 0 failures to 398 / 0. The existing tests missed all three bugs because they only cover perfect predictions, binary classification, and micro averaging on batches where every class is present.

### DIFF
--- a/deep/src/main/java/smile/deep/metric/Averaging.java
+++ b/deep/src/main/java/smile/deep/metric/Averaging.java
@@ -16,6 +16,8 @@
  */
 package smile.deep.metric;
 
+import smile.deep.tensor.Tensor;
+
 /**
  * The averaging strategy to aggregate binary performance metrics across
  * multi-classes.
@@ -43,5 +45,18 @@ public enum Averaging {
      * Weighted macro for imbalanced classes. Note that weighted recall is
      * equal to accuracy.
      */
-    Weighted
+    Weighted;
+
+    /**
+     * Returns the support-weighted average of per-class scores. Shared by the
+     * metrics so that they all reduce the {@link #Weighted} strategy the same
+     * way, including the empty-state case right after {@code reset()}.
+     * @param score the score of each class.
+     * @param size the number of samples of each class.
+     * @return the weighted average, or 0 if there is no sample.
+     */
+    static double weighted(Tensor score, Tensor size) {
+        double n = size.sum().doubleValue();
+        return n == 0.0 ? 0.0 : score.mul(size).sum().doubleValue() / n;
+    }
 }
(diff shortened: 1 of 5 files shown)


haifengl:
Thanks a lot!

PR #878, opened 2026-07-26
Fix tail p-values collapsing to 0 in Gamma/GLM, and NegativeBinomial pmf overflow

gaoflow:
`Gamma.regularizedUpperIncompleteGamma` returns `1 - (1 - Q)` on the `x >= s + 1` branch: the continued fraction evaluates `Q`, converts it to `P`, and the caller converts it back. The round trip through a number near 1 quantises `Q` to multiples of 2^-53 and floors it at exactly 0 once `Q < 2^-54`.

```
ChiSqTest.test({150,50,150,50}, {.25,.25,.25,.25})   chisq=100, df=3
    pvalue = 0.0                            (true 1.554e-21)
PoissonDistribution(500).cdf(300) = 0.0     (true 2.836e-22)
BoxTest.ljung(squared bitcoin log returns, 5), q=299.6
    pvalue = 0.0                            (true 1.217e-62)
```

That covers the three `ChiSqTest.test` overloads, both `BoxTest` statistics and `PoissonDistribution.cdf`. The fix returns the fraction's own `Q` and derives `P = 1 - Q`, as `gammq`/`gammp` do in Numerical Recipes; `P` is unchanged bit for bit.

A differential over the rest of the family against mpmath (3339 points across `Gamma`, `Beta`, `Erf` and the CDFs and tests built on them) found two more sites:

- `GLM.ztest()` built the two-sided p-value as `2.0 - Erf.erfc(-|z|/sqrt(2))`, i.e. `2 - (2 - erfc)`, so it floors at 0 for `|z|` over about 8.3. `GLMTest.testDefault` pins those two as `0.00000` at 1e-4; they are 5.0e-108 and 4.3e-135.
- `NegativeBinomialDistribution.p(k)` returns `Infinity` from k=162 and `NaN` from k=169 — only 2.4 sigma out at `r=10, p=0.1`. `logp` right below it already had the stable lgamma form.

`Beta.regularizedIncompleteBetaFunction`, `Erf.erfc`, `KSTest.qks`, `TDistribution.cdf`/`cdf2tailed` and the t-, F- and correlation tests were checked and are correct — they all evaluate the small quantity directly, which is what `Gamma` now does too. 283 of the 3339 points were wrong before, 0 after.

Tests added to the six affected test classes; all 7 fail on master. `:base:test` goes 2600 -> 2605 tests and `:core:test` 1312 -> 1313, with an identical set of pre-existing failures either way (native BLAS and ONNX on my machine).

### DIFF
--- a/base/src/main/java/smile/math/special/Gamma.java
+++ b/base/src/main/java/smile/math/special/Gamma.java
@@ -149,7 +149,7 @@ public class Gamma {
             igf = regularizedIncompleteGammaSeries(s, x);
         } else {
             // Continued fraction representation
-            igf = regularizedIncompleteGammaFraction(s, x);
+            igf = 1.0 - regularizedUpperIncompleteGammaFraction(s, x);
         }
 
         return igf;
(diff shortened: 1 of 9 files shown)


haifengl:
Thank you!

PR #879, opened 2026-07-28
Fix MatthewsCorrelation int overflow and zero-denominator NaN

gaoflow:
`MatthewsCorrelation.of` has two silent numeric defects.

**Int overflow flips the sign.** `tp * tn - fp * fn` is computed as `int` and wraps at 2^31. A 90%-accurate classifier on 110k samples (tp=tn=50000, fp=fn=5000) reports -0.6016 instead of +0.8182 — a wrong-sign metric, no exception. `AUC.of` in this same package already guards this exact case ("*for large sample size, overflow may happen for pos * neg. switch to double to prevent it.*"), but the guard was never carried over here. Now computed in `long`.

**Zero marginal returns NaN.** A constant predictor zeroes a confusion marginal, so the denominator is 0 and the method returns 0/0 = NaN. It now returns 0, matching scikit-learn's `matthews_corrcoef` and the standard convention.

Expected values come from `sklearn.metrics.matthews_corrcoef`; the currently-correct cases are bit-identical after the change (`long` and `int` agree there). Added tests cover the overflow sign-flip, the degenerate predictor, and the MCC in [-1, 1] invariant. The rest of `validation/metric` was checked for the same int-product overflow and is clean (AUC/AMI already accumulate in double).

### DIFF
--- a/core/src/main/java/smile/validation/metric/MatthewsCorrelation.java
+++ b/core/src/main/java/smile/validation/metric/MatthewsCorrelation.java
@@ -67,8 +67,14 @@ public class MatthewsCorrelation implements ClassificationMetric {
         int fp = matrix[0][1];
         int fn = matrix[1][0];
 
-        int numerator = (tp * tn - fp * fn);
+        // Compute in long to avoid int overflow of tp * tn on large samples (cf. AUC).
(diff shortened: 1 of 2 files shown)


haifengl:
Thank you!

PR #880, opened 2026-07-28
Fix int overflow in CorTest.spearman and CorTest.kendall for large samples

gaoflow:
Two int-overflow bugs in `CorTest` corrupt the correlation coefficient on ordinary-sized inputs.

**Spearman.** `CorTest.spearman` computes the rank normalization as

```java
double en3n = n * n * n - n;
```

`n` is `int`, so `n * n * n` is evaluated in 32-bit arithmetic and overflows before the widening to `double`. For n >= 1291 (1291^3 = 2,151,683,880 > Integer.MAX_VALUE) `en3n` wraps negative and the returned coefficient is garbage — it even leaves the mathematically required range [-1, 1]:

```
n=1000  rho = -1.0        (correct, no overflow)
n=1291  rho =  3.0078...  (should be -1.0)
n=2000  rho = 28.12...    (should be -1.0)
```

(x = 0..n-1, y reversed: strictly anti-correlated ranks, true Spearman rho = -1.) Fix: widen the first operand so the whole product promotes, `(double) n * n * n - n`. Note `(double)(n * n * n)` would not help — the int product still overflows before the cast.

**Kendall.** `CorTest.kendall` accumulates the concordant/discordant and tie counters `is`, `n1`, `n2` as `int`. They reach n(n-1)/2, overflowing for n >= 65537, after which `tau = is / (sqrt(n1) * sqrt(n2))` takes the square root of a wrapped-negative count and returns `NaN` (true tau = 1 for a perfectly concordant input). Widened to `long`.

This is the same widen-before-you-multiply issue already guarded in `AUC.of` ("switch to double to prevent it"); the pattern just wasn't carried over to `CorTest`.

Verified against scipy (`spearmanr` / `kendalltau`, exact closed form) and the oracle-free invariant that a correlation coefficient must lie in [-1, 1]. Added `testSpearmanLargeSample` (n = 1290/1291/2000) and `testKendallLargeSample` (n = 65537); both fail on master and pass with the fix. The existing small-sample tests are unchanged and stay bit-identical.

### DIFF
--- a/base/src/main/java/smile/stat/hypothesis/CorTest.java
+++ b/base/src/main/java/smile/stat/hypothesis/CorTest.java
@@ -166,7 +166,8 @@ public record CorTest(String method, double cor, double t, double df, double pva
             d += MathEx.pow2(wksp1[j] - wksp2[j]);
         }
 
-        double en3n = n * n * n - n;
+        // for large n, n * n * n overflows int before widening. promote first.
+        double en3n = (double) n * n * n - n;
         double fac = (1.0 - sf / en3n) * (1.0 - sg / en3n);
         double rs = (1.0 - (6.0 / en3n) * (d + (sf + sg) / 12.0)) / Math.sqrt(fac);
         fac = (rs + 1.0) * (1.0 - rs);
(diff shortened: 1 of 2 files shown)


haifengl:
Thank you!

PR #884, opened 2026-08-12
Studio: add null safety and fix `evalMagic` in code cells

TsybulkaM:
### Description
- **Bugfix / Guard:** Add `cli.agent() != null` check when AI-agents is inactive or disconnected.
- **Fix:** Move `evalMagic` parsing forward to prevent `SourceCodeAnalysis.CompletionInfo` from stripping `//!` commands as comments.


haifengl:
Thanks!

PR #885, opened 2026-08-14
fix: correct theme setting logic in SettingsDialog

TsybulkaM:
### Description:
- **Fix:** correct theme setting logic in SettingsDialog 
- **Feature:** enhance caret visibility handling in ThemedTextArea
- **Docs:** update README for latest SMILE package download instructio

### DIFF
--- a/studio/README.md
+++ b/studio/README.md
@@ -77,8 +77,9 @@ interacting with your data with natural language.
 
 1. **Download & Unzip**
 
-Download the [SMILE package](https://github.com/haifengl/smile/releases/download/v6.2.3/smile-6.2.3.zip)
-and unzip it on your machine.
+Download the latest SMILE package from the
+[latest release page](https://github.com/haifengl/smile/releases/latest)
+(grab the `smile-<version>.zip` asset) and unzip it on your machine.
 
 2. **Configure Your Environment**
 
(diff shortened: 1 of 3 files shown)


haifengl:
Thanks!

haifengl:
Besides caret visibility, this PR already changes caret update policy to `NEVER_UPDATE` for noneditable textarea. Can you please share the considerations for this change? IMHO, this actually downgrades the usability as I cannot easily see the training updates in Notebook output area and cannot follow the agent output in agent panel. If there is not strong justification, I would like to revert this new behavior. Thanks!

TsybulkaM:
I just would like to reduce amount of carets and think this change improve UX. Usually, a visible blinking caret universally signals that a field is interactive and ready for text input.

haifengl:
It is good to hide caret, which is achieved by
```
        var caret = new org.fife.ui.rtextarea.ConfigurableCaret() {
            @Override
            public void setVisible(boolean visible) {
                super.setVisible(visible && isEditable());
            }
        };
        caret.setBlinkRate(getCaret().getBlinkRate());
        setCaret(caret);
```

What concerns me is the below:
```
        syncCaretUpdatePolicy();
```
which sets caret update policy to `DefaultCaret.NEVER_UPDATE` that causes issues with autoscrolling for output panes.

TsybulkaM:
I agree. `syncCaretUpdatePolicy` can be reverted, if it downgrades the usability.

PR #890, opened 2026-08-28
feat: add OrcaRouter as a named AI provider in SMILE Studio

dulcestentaciones2920-debug:
Adding [OrcaRouter](https://www.orcarouter.ai) as a named AI provider in SMILE Studio means Studio users get the gateway's full stack directly from **File > Settings > AI Service**, instead of treating it as an anonymous "Chat Completions Compatible" base URL.

OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding orcarouter as a first-class provider means this project's users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL.

The change mirrors the existing **OpenAI** entry in `SettingsDialog` and `SmileStudio.createLLM()`. OrcaRouter speaks the OpenAI Chat Completions protocol, so it reuses the `OpenAI` client pointed at `https://api.orcarouter.ai/v1` with an `orcarouter/*` routing model. The Settings dialog prefills that base URL and suggests `orcarouter/fusion-mini`, `orcarouter/fusion-flash`, `orcarouter/fusion`, and `orcarouter/free` as model choices.

It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

**Files changed**
- `studio/src/main/java/smile/studio/SettingsDialog.java` — new `OrcaRouter` service option, key, base URL default, and model suggestions
- `studio/src/main/java/smile/studio/SmileStudio.java` — new `case "OrcaRouter"` in `createLLM()`, mirroring the `OpenAI` wiring
- `studio/README.md` — provider row in the Settings table

**Verification**
- The full studio module compiles cleanly against the release dependency set.
- Live-tested the new provider path (`new OpenAI("orcarouter/fusion-mini").withApiKey(...).withBaseUrl("https://api.orcarouter.ai/v1")`) with a real chat completion through the OrcaRouter endpoint — returned a valid response.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/studio/README.md
+++ b/studio/README.md
@@ -561,6 +561,7 @@ Open via **File > Settings…**. Choose an AI service provider from the drop-dow
 | **Anthropic** | Claude models; set API key, optional base URL, model |
 | **Google Gemini** | Gemini models via native API; set API key, model |
 | **Google Vertex AI** | Gemini via Vertex; set API key, base URL, model |
+| **OrcaRouter** | OpenAI-compatible AI gateway; set API key, base URL (defaults to `https://api.orcarouter.ai/v1`), model |
 
 All fields (API key, base URL, model) are stored in Java `Preferences` (OS keychain / registry). API keys set in Settings take precedence over environment variables.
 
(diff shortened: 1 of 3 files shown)


haifengl:
Thanks for contributing! Instead of adding new AI services, I change Base URL to a combo box with OrcaRouter and other command local inference options. Please try it with master branch.

